PR #1, opened 2021-07-26
Draft design of indexer-provider ingestion protocol.

adlrocha:
[Rendered](https://github.com/filecoin-project/indexer-reference-provider/blob/docs/design/docs/design.md)


adlrocha:
@gammazero, @willscott, would you mind giving a first review before sharing it with other folks? Thanks!

docs/design.md
```diff
@@ -0,0 +1,257 @@
...
+```
+- `ID` is the unique identifier used by the data provider to identify this update.
+- `Previous` links the current advertisement for the provider with the previous one. This allows
+providers to track the announcements they are broadcasting, and 
+indexer nodes to realize when they have missed a previous advertisement.
+- `Provider` determines the source of the advertisement.
+- `Signature` is a signature over `ID` and `Previous` to authenticate that the advertisement has been generated by `Provider`.
+- `GraphSupport` is used by providers to signal indexer nodes that they support IPLD-aware ingestion sessions (see [protocol](#protocol)).
+
+`AdvertisementID` can be chosen by data providers in the way that suit them best according to their internal
```
willscott, line 42: can we make this a hash of the advertisement, rather than an opaque identifier so that it self-certifies the content (e.g that the `Signature` here ends up validating the contents of the advertisement)

docs/design.md
```diff
@@ -0,0 +1,257 @@
...
+```go
+type IngestionRequest struct {
+        ID              AdvertisementID
+        Start           uint64
+        Count           uint64
+}
+```
+- `ID`: `AdvertisementID` for which data is being requested.
+- `Start`: Starting entry from which data wants to be requested.
+- `Count`: Maximum number of entries that want to be received in the response.
```
willscott, line 73: is this for pagination within the advertisement?
The indexer won't know the total count though, so how does it know when it's gotten all?
Isn't the common case 'i want all of the advertisement'? can we defer pagination for now?

docs/design.md
```diff
@@ -0,0 +1,257 @@
...
+....
+### Ingestion Response
+Data providers use `IngestionResponse`s to answer to ingestion requests.
+
+```go
+type IngestionResponse struct {
```
willscott, line 79: note that unless signature is over hash of overall advertisement there's no authenticity guarantee over the actual cid payloads here,
so a provider could offer an advertisement but map it to different sets of CIDs to different index nodes.

docs/design.md
adlrocha: Yes. This is for pagination within the advertisement. `Count` is only used by the indexer node to limit the maximum amount of entries it wants to receive in a single Response.
The provider signals the total number of entries of the advertisement in `Size`, so after the first Response, the indexer gets knowledge of the total number of entries of the advertisement. 

I personally think we should aim to include pagination in the MVP, because the amount of data to be exchanged between provider an indexer may be quite large even for the MVP, and we can easily reach the `network.MessageSizeMax` soft limit if we choose to implement a message-based protocol (but happy to be convinced otherwise :) )

docs/design.md
adlrocha: Great point. I added the opaque identifier as an attempt to make the protocol as general possible, but I am relax the self-certification requirement. I'll revise this.

docs/design.md
willscott: i'd like to have the concrete transport / protocol considered jointly with this if possible. I feel like there's a range of options and am not convinced pagination like this will be easiest to implement.

If i'm doing a bulk HTTP-style upload, I probably don't care too much about size, the http layer will let me push the whole list of CIDs in one go, and potentially do range queries over the record if partial gaps need to be filled. Having it be non-dynamic means the provider can generate the 'announcement' once and leave it as a static json file on s3 or similar, rather than needing a dynamic endpoint to handle potentially-dynamic paging.

If i'm doing a libp2p-style protocol, we have bitswap and graphsync already for transferring dags. can't we use one of those rather than re-inventing a chunking method here?

Maybe the underlying thing to figure out is: I think it's never going to the case that I want to partially fetch an announcement. in order to verify it, i'm either going to want all of it, or none of it as the granularity that i do rate limiting on as an indexer node.

adlrocha: > Having it be non-dynamic means the provider can generate the 'announcement' once and leave it as a static json file on s3 or similar, rather than needing a dynamic endpoint to handle potentially-dynamic paging.

This should definitely be a requirement. Announcement should be static. And ideally the ingestion protocol should be stateless to allow any kind of use case.

> If i'm doing a libp2p-style protocol, we have bitswap and graphsync already for transferring dags. can't we use one of those rather than re-inventing a chunking method here?

Agree! As I mention by the end, ideally if the provider represent data internally using IPLD, we may default to Graphsync, but I wasn't sure if we could expect this for all providers, specially because for the initial design we were considering the use of pagination already. 

> I think it's never going to the case that I want to partially fetch an announcement. in order to verify it, i'm either going to want all of it, or none of it as the granularity that i do rate limiting on as an indexer node.

Right. Ingestions are going to be atomic. You'll want to fetch all or nothing.

Thanks for the feedback. I'll do another pass addressing this.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+providers to track the announcements they are broadcasting, and 
+indexer nodes to realize when they have missed a previous advertisement.
+- `Provider` determines the source of the advertisement.
+- `Signature` is a signature over `Advertisement.ID` to authenticate that the advertisement has been generated by `Provider`.
+- `GraphSupport` is used by providers to signal indexer nodes that they support IPLD-aware ingestion sessions (see [protocol](#protocol)).
+
+Indexer nodes use these IDs in their ingestion requests to fetch indexing
+data from the provider.
+
+For instance, a Filecoin miner may include as part of a new `Advertisement` the set of CIDs included in a new deal, and it may
```
gammazero, line 52: I do not think the Advertisement should ever include CIDs.  This should only be something that is fetched by the indexer.  There are a number of reasons for this:
- Too much data for pubsub
- Need response pagination
- Indexer should get data only when it determines it is time for an update
- Allows for batching and for skipped updates

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+amount of CIDs.
+
+Once all the data for an indexer has been fetched, the indexer can use the `Advertisement` and the list of entries received to authenticate the
+ingestion by computing the Cid of the entries and verifying that it is equal to `IndexID`, and then verifying the signature
+of the `Advertisement`.
+
+If the full sync is performed successfully, the indexer node updates the latest AdvertisementID seen for `p` to latest. 
+
+Indexer nodes run ingestion sessions for different providers in parallel, so an indexer node can sync with several provideres in parallel.
+They can also set a `prov_parallel_ingestion` config to determine if it wants to support parallel ingestions for
```
gammazero, line 269: I do not think we need `prov_parallel_ingestion`.  Since the diffs of a single provider need to be applied in order, it would mean that the indexer will have to temporarily store all the data for all updates except the oldest one, and this is not something we should do.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+and repeats the process over and over again until the `AdvertisementID` seen in `Previous` for the latest response
+equals the latest one seen by indexer node for that data provider and the full sync for the missing range is complete.
+
+Note that pagination is performed at an `Entry` level, i.e. there is no intra-Entry pagination (at least not initially). The data provider should be responsible for dividing single Entries with a large number of CIDs into
+smaller entries when computing the `Size` of the data indexed in `AdvertisementID`. If we identify this as too strong
+of a requirement, we can also introduce intra-Entry pagination to support seamless exchange of entries with a large
+amount of CIDs.
+
+Once all the data for an indexer has been fetched, the indexer can use the `Advertisement` and the list of entries received to authenticate the
+ingestion by computing the Cid of the entries and verifying that it is equal to `IndexID`, and then verifying the signature
```
gammazero, line 263: A quick note on how the CID of the entries is computed would be good.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+                Previous: "latest-1",
+                IndexerID: Link_Index,
+                Provider: p,
+                Signature: []byte{},  // The original signature of the advertisement.
+                GraphSupport: false, 
+        }
+        Entries: {
+                []Entry{
+                        Cids: {Cid1, Cid2, ...},
+                        Metadata: []byte{},
```
gammazero, line 247: Look at the [design doc](https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825#a9b0aa68c1374ab4897aac7917e7262f) to make sure these align.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+                }
+}
+```
+- If `Size > len(Entries)`, the indexer-node sends a new request to keep fetching data for `latest` with `Start: 1`
+until all entries for the Advertisement have been received. It then inspects `Advertisement` to see if `Previous` is equal
+to the latest `AdvertisementID` seen for that provider, if this is not the case, it sends a new request with `latest-1`
+and repeats the process over and over again until the `AdvertisementID` seen in `Previous` for the latest response
+equals the latest one seen by indexer node for that data provider and the full sync for the missing range is complete.
+
+Note that pagination is performed at an `Entry` level, i.e. there is no intra-Entry pagination (at least not initially). The data provider should be responsible for dividing single Entries with a large number of CIDs into
```
gammazero, line 257: The design doc is very specific about how [response pagination](https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825#278589e9c95d4363a5986b4309fc0f5a) works.  Maybe consider using  that text, or something equivalent.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+                GraphSupport: false, 
+        }
+        Entries: {
+                []Entry{
+                        Cids: {Cid1, Cid2, ...},
+                        Metadata: []byte{},
+                }
+}
+```
+- If `Size > len(Entries)`, the indexer-node sends a new request to keep fetching data for `latest` with `Start: 1`
```
gammazero, line 251: The value of `Start` has to do with pagination, so that should be discussed in the next paragraph.  This paragraph is discussing chained responses.  See the [Chained Responses](https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825#ba49552aa0bd47d192317a9f34e94555) section of the design doc for reference to make sure these match.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+push Advertisements and immediate ingestions (see [protocol](#protocol)).
+
+## Protocol
+The interaction protocol between data providers and indexer nodes can be divided in two subprotocols: the advertisement protocol, and the ingestion protocol.
+
+## Advertisement Protocol
+It orchestrates the announcement of udpates from data providers to indexer nodes. Every time a data provider receives
+an update for the data they provide (e.g. new deal in a Filecoin miner), they may choose to broadcast an `Advertisement`
+to the network so indexer nodes can start indexing this data. Three kinds of updates will initially be supported:
+- `Pull` Advertisements: Data providers broadcast the `Advertisement` for the new data using a shared
```
gammazero, line 155: Pull should be split into two types:
- **Subscribe**: Indexer receives advertisement on gossip pib-sub channel
- **Poll**: Indexer periodically requests latest advertisement from provider

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Data Structures
+Data providers and indexer nodes can use any internal representation they want for indexed data. They only
+data structures enforced by the protocol are in the interaction between the two.
+
+### Advertisements
+Advertisements are used by data providers to share updates about the data they are providing. Ideally,
+these advertisements are published in a shared pub-sub channel between providers and indexer nodes. However,
+some indexer nodes may choose to provide a dedicated endpoint to allow certain authenticated providers to push announcements and updates directly to them.
+- `Pull` Advertisements: Indexer-nodes pull advertisement from a shared pub-sub channel.
```
gammazero, line 21: `Pull` IMO is not the correct description, because the indexer is not initiating the action.  I would call this `Subscribe` if looking at it from the indexer side, or `Publish` if looking at it from the provider side.

There is an additional mechanism where the indexer does initiate action and asks the provider directly for the latest advertisement.  I would call that `Poll` or `Pull`.  This is discussed [here](https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825#edfdbffea3d1407ba7ac696d490f0b3a) in the design doc.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+data provider.
+- `Push` Advertisements: Some indexer nodes may support advertisement pushes by data providers. Indexers 
+will trigger a new ingestion session to gather the data for that update. 
+- `Immediate` Ingestion: Some data providers may want to share updates for a small number of CIDs which doesn't
+require the generation of a dedicated `Advertisement` and triggering an ingestion session. For this case, indexer
+nodes will provide an endpoint for providers to push small updates directly to them. Indexer nodes may
+restrict immediate ingestion updates to authenticated data providers __(we3.storage case)__.
+
+## Ingestion Protocol
+When indexer nodes realize that they are not in sync with a data provider they are indexing data for, they trigger
```
gammazero, line 166: Minor suggestion:  Be specific about how an indexer knows it is not in sync.

"When an indexer sees that a provider has advertised a later index than the indexer currently has..."

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+new ingestion sessions with them to gather the new updates. This is orchestrated by the ingestion protocol.
```
gammazero, line 167: Replace the first sentence here:
> When indexer nodes realize ... gather the new updates.

With the paragraph below.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+Indexer nodes track locally the latest `AdvertisementID` seen for each data provider. When an `Advertisement` for an unseen `AdvertisementID` is seen for a data provider, a new ingestion session is triggered with the data provider.
+
+Data providers are responsible for broadcasting `Advertisement`s and generating their `AdvertisementID`s in the way that
+suits them best. This means that they may choose to use IPLD data structures under the hood to link their advertisements
+with their indexed data, so that `AdvertisementID` is a link to the indexed data for that `Advertisement`. If this is
+the case, the provider may set the `GraphSupport` flag in its `Advertisement` to signal indexer nodes that they
+can use IPLD-aware data transfer protocol, such as Graphsync, for the ingestion session.
```
gammazero, line 175: +1

gammazero, line 175: 👍

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+According to how the data providers choose to represent their indexed data locally, two flavors of the ingestion protocols can be instantiated by the indexer node.
+
+### IPLD-aware Ingestion
+Initially, this will be default transport. If a data provider is indexing data in a way that `AdvertisementID` is linked with the indexed data using IPLD, the `GraphSupport` flag is enabled in its Advertisements to signal indexer nodes that they can use IPLD-aware data transfer
+protocols.
+
+When an indexer node sees an unseen advertisement `latest`, it will trigger an ingestion session to sync with the
+data provider. If `GraphSupport` is enabled, the indexer node can use Graphsync or `go-data-transfer` to fetch `latest` DAG, which links to the data that needs to be ingested, using a [`ExploreRecursiveEdge`](https://ipld.io/specs/selectors/) selector. `latest` will include a link to the data structure for the previous `AdvertisementID` to latest, so Graphsync is able to fetch all the chain of updates from the provider seamlessly up to the latest one seen by the indexer node.
+
```
gammazero, line 189: Will the new dag ever contain CIDs that the indexer already knows about and would be fetching unnecessarily?

Does the indexer need to know how to generate the difference between the two dags so that it knows which CIDs to remove?

gammazero (commented):
This is a great start.  I left a number of comments and questions.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+## Data Structures
+Data providers and indexer nodes can use any internal representation they want for indexed data. They only
+data structures enforced by the protocol are in the interaction between the two.
+
+### Advertisements
+Advertisements are used by data providers to share updates about the data they are providing. Ideally,
+these advertisements are published in a shared pub-sub channel between providers and indexer nodes. However,
+some indexer nodes may choose to provide a dedicated endpoint to allow certain authenticated providers to push announcements and updates directly to them.
+- `Pull` Advertisements: Indexer-nodes pull advertisement from a shared pub-sub channel.
+- `Push` Advertisements: Authenticated providers can push advertisements to indexer-nodes.
```
gammazero, line 22: Does a miner have to be discovered first, or is the push supposed to be another mechanism for provider discovery?

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+advertisement for a data provider, they trigger a new ingestion session to sync with the provider. 
+
+## Data Structures
+Data providers and indexer nodes can use any internal representation they want for indexed data. They only
+data structures enforced by the protocol are in the interaction between the two.
+
+### Advertisements
+Advertisements are used by data providers to share updates about the data they are providing. Ideally,
+these advertisements are published in a shared pub-sub channel between providers and indexer nodes. However,
+some indexer nodes may choose to provide a dedicated endpoint to allow certain authenticated providers to push announcements and updates directly to them.
```
gammazero, line 20: Are announcements distinct data structures that should be described?  It is not clear to me what an announcement is vs. an advertisement.

> and updates directly

Change to: "and updates to individual CID index values"

docs/design.md
gammazero, line 21: Need to add the method where we read the advertisements from an S3 bucket, or a file, IPFS, or whatever.  I think we can call it the `Load` method?  Unless you think of something better.  Not sure if that goes in this section of the doc or in the "Advertisement Protocol" section, or both.

docs/design.md
```diff
@@ -0,0 +1,279 @@
...
+- `Pull` Advertisements: Data providers broadcast the `Advertisement` for the new data using a shared
+pub-sub channel. When indexer nodes see these updates, they trigger an ingestion session to sync with the
+data provider.
+- `Push` Advertisements: Some indexer nodes may support advertisement pushes by data providers. Indexers 
+will trigger a new ingestion session to gather the data for that update. 
+- `Immediate` Ingestion: Some data providers may want to share updates for a small number of CIDs which doesn't
+require the generation of a dedicated `Advertisement` and triggering an ingestion session. For this case, indexer
+nodes will provide an endpoint for providers to push small updates directly to them. Indexer nodes may
+restrict immediate ingestion updates to authenticated data providers __(we3.storage case)__.
+
```
gammazero, line 164: Need to add `Load` method mentioned in earlier comment.

docs/design.md
adlrocha: This is a misunderstanding, I don't expect Advertisements to include all CIDs for a deal, but to reference a specific set of CIDs. It is probably not described correctly, I'll rephrase it.

docs/design.md
```diff
@@ -0,0 +1,279 @@
+# Indexer Reference Provider Design
+The goal of this reference implementation is to flesh out all the protocol details
+for data providers to be able to interact with indexer nodes. 
+
+## Overview
+- Data providers keep a local index of all the data they are providing.
```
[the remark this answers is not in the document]
adlrocha: You mean conceptually or to define its format? Adding some clarification in the next PR, let me know if it still needs something else.

docs/design.md
adlrocha: You are right, parallel ingestion probably works for subsequents adds, but the moment an update arrives for the same CID in the wrong order things break (this would make sense if we used CRDTs which is not the case). I'lll remove this part.

docs/design.md
adlrocha: A few thoughts regarding `Entries` and and `Requests`:
- Do we really need to add `prev_index` in requests as described in the design doc? If we are building a message-oriented protocol and we want the server side to be stateless, it is better to tailor requests of indices as we need them. For Graphsync, we will just send a selector for the latest advertisement and terminate when we reach our current state.

- Regarding Entries, do we expect `Cids` and `RmCids` to be set in the same update? For Filecoin miners I don't think so, but for other use cases probably yes. I'll update it accordingly.

docs/design.md
adlrocha: That paragraph was supposed to describe how pagination +  chaining worked combined (but I think it failed :) ). I changed it to describe one after the other as it is in the design doc.

docs/design.md
adlrocha: You're totally right. Adding it.

docs/design.md
adlrocha: Good catch, I was rephrasing and forgot to remove it. Thanks!

docs/design.md
adlrocha: Good catch, I was missing this. I added the new functions to the interface. Let me know what you think.

docs/design.md
adlrocha: My reasoning here was that some indexers may be open to: 
- Accept new providers that proactively push advertisements to them (after the corresponding verification); 
- Accept advertisement pushes from already verified providers (because they are a priority?) so it these indexers don't have to wait to catch the advertisement from the pubsub channel.
It would be up to the indexer to config the policies it supports.

docs/design.md
adlrocha: I am using them interchangeable. I'll review the full text to make it as consistent as possible.

docs/design.md
adlrocha: Agree. Fixed!

docs/design.md
adlrocha: I think it is up to the provider to decide how it wants to link advertisements to entries, and how to provide the entries for ingestion.

For the Filecoin case this is quite straightforward, because Advertisements will probably be linked to new or expired deals, and deals that share CIDs will represent a metadata update in the index. 

I would expect duplicate CIDs on independent advertisements. The indexer should just disregard those that already has (which is what it already does when calling `Put` of `indexer-core`). 

On a related note, something we will need to figure out is how to perform snapshots of provider indexers to prune chains that are becoming too long. I'll add a placeholder in the doc.

docs/design.md
gammazero: The indexer looks at the previous entry to see if already has the previous change set, or if it needs to fetch that first (since differences are applied in order).  Without `prev_index` the indexer will not know whether it already has the previous index or whether it missed some updates and needs to fetch them.  When getting previous indexes, the indexer knows when to stop when it sees a `prev_index` that it already has.

> For Graphsync, we will just send a selector for the latest advertisement and terminate when we reach our current state.

How does the indexer know if the selector only needs the latest, or if it needs the last 3 because some were missed?

CIDs to add and CIDs to remove can be in the same entry or in different entries.  It is up to the provider how they want to break it up.  I can see reasons why a provider might want to do one or the other (depending on how their diff calculation works), so I figured we should allow both.
Maybe I am not understanding how diffs work with IPLD, but to get the correct set of diffs as described in the design doc,  the advertisement will need to include the previous index ID.

docs/design.md
gammazero: OK, then somewhere in the doc state that a `Push` can also serve as a discovery mechanism.

docs/design.md
willscott: @hannahhoward suggested it would not be hard to add an extension to graphsync to include a 'skip subdag' extension to allow the indexer in its request to provide its most recent synced link as an indication that it has already received objects from that point back and to stop traversal at that link.

docs/design.md
gammazero: > I would expect duplicate CIDs on independent advertisements. The indexer should just disregard those that already has

Even if the CIDs are the same as previously seen, since they are part of a new deal they will almost certainly have different metadata.  This means that that the indexer will store multiple values for the same CID and provider.

I guess this is OK, and even good in that it allows the providers an easy way to delete CIDs that are no longer used.  When a deal expires, the provider tells the indexer to delete all the CIDs that have the metadata for that deal. If the CIDs were used in other deals, then the indexer will still have those values for the CIDs. The CID will only disappear from the indexer when it has no more values.

If the above sounds correct, then I will describe that in the design document. It should also be described clearly here to inform providers how to delete stale CIDs from the indexer without having to maintain state info (ref counts) for every CID in all their deals.

docs/design.md
adlrocha: > The indexer looks at the previous entry to see if already has the previous change set, or if it needs to fetch that first (since differences are applied in order). Without prev_index the indexer will not know whether it already has the previous index or whether it missed some updates and needs to fetch them.

Of course, advertisements should include a pointer to the previous advertisement in the chain. What I meant is that for ingestion requests, there is no need for `prev_index` (as suggested in the design doc) as I think the one keeping the state of what to request next should be the indexer and not the provider.

From what I understand:
- Providers are responsible for keeping a local index of the content they are providing, and linking updates over the index to advertisements. Advertisements are linked sequentially to each other building a chain. The provider is the one responsible for linking Advertisement to the set of new and removed CIDs in their index however it wants (storing a static IPLD structure, computing this relationship in every request, etc.). The only requirements is that advertisements should be linked to entries in a way that they can be authenticated when ingested by the indexer.
- Indexers receiving an unseen advertisement are responsible for following the links and gathering the updates up to their last sync. `latest_ad --> entries --> previous_ad --> entries --> ... --> last_sync`.
- If the provider supports graphsync it should build its internal data structures and the links between advertisement and entries using a DAG structure so Graphsync can seamlessly do the traversal from advertisement to linked entries, to previous advertisement, to linked entries, and so on. For the non-graphsync case, the indexer will perform requests for each advertisement ID and proactively traverse the chain by making stateless requests to the provider for entries related to an advertisement until it fully syncs.

Adding a quick figure to support my explanation (in case it helps): 
image

Does this make sense to you or am I missing something?

gammazero: Ok, I understand what you are saying.  Appreciate explanation.  Diagram is great to include in this doc aslo.

gammazero (approved):
So far this looks good.  Not sure if it is better to keep this as a PR until there is more feedback, of merge now.

adlrocha:
> So far this looks good. Not sure if it is better to keep this as a PR until there is more feedback, of merge now.

 I think it's easier to gather feedback in the PR. @willscott, thoughts?

gammazero (approved):

adlrocha:
hey, @mikeal. This a draft spec of the ingestion protocol we are considering between indexer and providers. We tried to bear in mind all of the requirements you shared with us, but can you have a quick look to give us feedback and see if what we are proposing makes sense. Thanks!

docs/design.md
masih:
```suggestion
ingestion by computing the CID of the entries and verifying that it is equal to `IndexID`, and then verifying the signature
```

docs/design.md
```diff
@@ -0,0 +1,333 @@
...
+If `Size > len(Entries)`, the indexer-node sends a new request to keep fetching data for `latest` with `Start: 1`
+until all entries for the Advertisement have been received. It then inspects `Advertisement` to see if `Previous` is equal
+to the latest `AdvertisementID` seen for that provider, if this is not the case, it sends a new request with `latest-1`
+and repeats the process over and over again until the `AdvertisementID` seen in `Previous` for the latest response
+equals the latest one seen by indexer node for that data provider and the full sync for the missing range is complete.
+-->
+
+Once all the data for an indexer has been fetched, the indexer can use the `Advertisement` and the list of entries received to authenticate the
+ingestion by computing the Cid of the entries and verifying that it is equal to `IndexID`, and then verifying the signature
+of the `Advertisement`. The Cid is computed by serializing the content and computing the IPLD link (see [linkingSystem](https://github.com/ipld/go-ipld-prime/blob/master/linking.go))
```
masih, line 304:
```suggestion
of the `Advertisement`. The CID is computed by serializing the content and computing the IPLD link (see [linkingSystem](https://github.com/ipld/go-ipld-prime/blob/master/linking.go))
```

docs/design.md
```diff
@@ -0,0 +1,333 @@
...
+
+If the full sync is performed successfully, the indexer node updates the latest AdvertisementID seen for `p` to latest.
```
masih, line 306:
```suggestion
If the full sync is performed successfully, the indexer node updates the latest `AdvertisementID` seen for `p` to `latest`. 
```

docs/design.md
```diff
@@ -0,0 +1,333 @@
...
+tracks `Sync` sessions, and new Advertisements trigger new ingestion sessions with the provider for a non-overlapping
+range between the current syncs and the new one. Thus, if there is a ingestion session with `latest = ID1` and
+a new advertisement arrives with `AdvertisementID: ID2` and `prov_parallel_ingestion > 0` then a new ingestion
+session is started for the range `(ID1, ID2] with `Ingest(ctx, p, ID1, ID2)`.
+-->
+
+## Provider Snapshots
+> Work in progress. Needs discussion. Do we need this?
+
+- If provider advertisement chains are becoming too large, they may choose to create a snapshot to
```
masih, line 323:
```suggestion
- If advertisement chains become too large, a provider may choose to create a snapshot to
```

docs/design.md
```diff
@@ -0,0 +1,333 @@
...
+                []Entry{
+                        Cids: {Cid1, Cid2, ...},
+                        RmCids: {Cid3, Cid4, ...},
+                        Metadata: []byte{},
+                }
+}
+```
+
+
+The response may be paginated either by the client requesting a maximum number of entries or by the server delivering up to a configured maximum number of entries.  To get the remaining entries, a subsequent request is made with its start set to the number of entries from the previous response.  So, if totalEntries equals 100 and the response contained 50 entries, the follow-up request should specify 50 for start to get the remaining entries.  This continues until the indexer has the complete response.
```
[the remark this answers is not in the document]
adlrocha: Good call, thanks @masih

masih:
Closing; to be replaced by:
* https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825
* https://www.notion.so/protocollabs/Indexer-Ingestion-Interface-4a120c698b31417385204ec401137cb1

PR #7, opened 2021-08-20
Interface changes and publication functions implementation

adlrocha:
_Do not merge. Adding it as a draft to give everyone visibility of what I am working on.
This PR depends on https://github.com/filecoin-project/storetheindex/pull/42_

The PR:
- Includes interface changes.
- Implements some of the functions of the interface to be able to generate new indeex and advertisements, and publish them to the gossipsub topic.

Pending:
- Tests


core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
-	ipldIngest legs.LegPublisher
+	lp legs.LegPublisher
 	// pubsubtopic where the provider will push advertisements
 	pubSubTopic string
+
+	// Local indexer with all provided data.
+	// NOTE: It may be the case that providers want to keep a local index with all
+	// provider data, and they want to store it with go-indexer-core. In this case
+	// we may need to wrap go-indexer-core so linkSystems can use it.
+	//indexer indexer.Interface
```
willscott, line 51: will we be able to efficiently access individual updates that have been indexed into go-indexer-core? I was expecting that would be lossy (or at least inefficient to extract back out the reverse index), and we'd want/need to have the cids split into individual updates stored separately.

core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
+	latestAdvID, err := e.getLatestAdv()
+	if err != nil {
+		return err
+	}
+	latestIndexLink, err := e.getLatestIndexLink()
+	if err != nil {
+		return err
+	}
+	// Lsys will store the index conveniently here.
+	_, indexLnk, err := schema.NewSingleEntryIndex(e.lsys, cids, nil, metadata, latestIndexLink)
```
willscott, line 143: how hard will it be to not store the cids themselves?
we could imagine that we may want to assume we can regenerate them on demand from the car, rather than having them duplicated into the datastore here, since they'll be the bulk of the data

core/engine/engine_test.go
```diff
@@ -0,0 +1,7 @@
+package engine
+
+import "testing"
+
+func TestHelloWorld(t *testing.T) {
```
willscott, line 5: ℹ️ remove or implement

core/engine/linksystem.go
```diff
@@ -0,0 +1,89 @@
...
+			return nil, err
+		}
+		return bytes.NewBuffer(val), nil
+	}
+	lsys.StorageWriteOpener = func(lctx ipld.LinkContext) (io.Writer, ipld.BlockWriteCommitter, error) {
+		buf := bytes.NewBuffer(nil)
+		return buf, func(lnk ipld.Link) error {
+			return ds.Put(datastore.NewKey(storageKey(lctx, lnk)), buf.Bytes())
+		}, nil
+	}
```
willscott, line 37: @warpfork consider if there's a way to avoid this boilerplate?

core/engine/engine_test.go
adlrocha: It's a placeholder to remind me that I need to implement tests, as I don't think I'll have time to jump into it today, that is why many things are in a "work in progress" state. I wanted to give others visibility early to iterate on the implementation.

core/engine/linksystem.go
adlrocha: I don't think we can (or want) to avoid this boilerplate. Each link system may implement a different interaction with the underlying datastore. In this case, we add a small prefix according to the schema type to identify to which schema type a CID belongs to.

core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
 
 var log = logging.Logger("reference-provider")
 
 var _ core.Interface = &Engine{}
 
+const (
+	// NOTE: For now this is an IPLD-aware reference provider, wo
+	// we need to force GraphSupport. This may change in the future
+	// and we may pass this as an option.
+	graphSupport = true
```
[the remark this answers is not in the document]
adlrocha: I don't know if I understand your comment. This is just to add the `graphSupport` flag to Advertisements. This is used to signal indexers if they can do IPLD-aware ingestion with the provider. 

For the MVP, `graphSupport` is always true, but in the future some providers may support alternative ingestion schemes like the paginated one, right? If this is not the case, we can remove this flag from here and the advertisement scheme right away.

core/engine/engine.go
adlrocha: This was actually the placeholder to ask you this exact question. I left this in case we want to support individual indexing from the reference provider, if this is not the case this can be removed. We only keep advertisements and updates in the datastore.

core/engine/engine.go
adlrocha: Great point. I was deferring this to the future. But I'll look into it right now.

core/interface.go
```diff
@@ -12,28 +11,40 @@ import (
...
+
+	// PutEvent notifies that new data has been added to the provider
+	PutEvent(ctx context.Context, cids []cid.Cid, metadata []byte) error
 
-	// Put new content in provider's local index. Putting new content
-	// in the index triggers the generation of a new advertisement.
-	Put(cids []cid.Cid, val indexer.Value) (bool, error)
+	// RemoveEvent sends an event to the reference provider to notify
+	// that new data has been removed from the provider
+	RemoveEvent(ctx context.Context, cids []cid.Cid, metadata []byte) error
```
masih, line 47: Naming seems inconsistent with other functions. in this interface
Without docs, I would have thought this functions removes an event.

Ditto re `PutEvent`

maybe something like:

```suggestion
	NotifyRemoved(ctx context.Context, cids []cid.Cid, metadata []byte) error
```

Also provider design doc need to be updated though its PR is not merged yet :)

core/interface.go
```diff
@@ -12,28 +11,40 @@ import (
...
 
 	// Push an update for a single entry.
 	// This can be used to perform updates for a small number of CIDs
 	// When a full advertisement is not worth it (web3.storage case).
 	// Indexer may only accept pushes from authenticated providers.
-	Push(ctx context.Context, indexer peer.ID, cid cid.Cid, val indexer.Value)
+	Push(ctx context.Context, indexer peer.ID, cid cid.Cid, metadata []byte)
+
+	// PutEvent notifies that new data has been added to the provider
+	PutEvent(ctx context.Context, cids []cid.Cid, metadata []byte) error
```
masih, line 43:
```suggestion
	NotifyPut(ctx context.Context, cids []cid.Cid, metadata []byte) error
```

core/interface.go
```diff
@@ -12,28 +11,40 @@ import (
...
+type Index schema.Index
+
+// IndexLink type
+type IndexLink schema.Link_Index
+
 // Interface for a reference provider
 type Interface interface {
 	// PublishLocal provides a new advertisement locally.
-	PublishLocal(ctx context.Context, ad Advertisement) error
+	PublishLocal(ctx context.Context, adv Advertisement) (cid.Cid, error)
```
masih, line 26: Would it make sense to update the documentation with what the returned CID represents?

core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
-	lsys := cidlink.DefaultLinkSystem()
-	lsys.StorageReadOpener = func(lctx ipld.LinkContext, lnk ipld.Link) (io.Reader, error) {
-		c := lnk.(cidlink.Link).Cid
-		val, err := ds.Get(datastore.NewKey(c.String()))
-		if err != nil {
-			return nil, err
-		}
-		return bytes.NewBuffer(val), nil
+func (e *Engine) PublishLocal(ctx context.Context, adv core.Advertisement) (cid.Cid, error) {
+	return e.publishLocal(ctx, adv)
```
masih, line 82: I am curious what's the rationale for this refactor? func signatures are identical.

core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
+	// provider data, and they want to store it with go-indexer-core. In this case
+	// we may need to wrap go-indexer-core so linkSystems can use it.
+	//indexer indexer.Interface
 }
 
 // New creates a reference provider engine with the corresponding config.
-func New(ctx context.Context, host host.Host, ds datastore.Batching, index indexer.Interface, pubSubTopic string) (*Engine, error) {
+func New(ctx context.Context,
+	cfg config.Config, host host.Host, ds datastore.Batching, index indexer.Interface,
+	pubSubTopic string) (*Engine, error) {
```
masih, line 57: Would it make sense to combine some of these arguments into config?

core/engine/engine.go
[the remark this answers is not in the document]
willscott: I think the main thing i was thinking was: if you're making the advertisement over gossipsub already, you probably support graphSupport for fetching. I'd expect the other form of advertisement/fetch to be sufficiently different that it wouldn't use this same engine / reference provider at all.
I'm not strongly opposed to using the same data format between the two and keeping this field though.

core/engine/engine.go
willscott: i think we don't need to support individual indexing for now

core/interface.go
adlrocha: Totally agree. Thanks @masih. I like this naming. 

I am afraid that several things may change while we develop the reference provider, so I'll defer updating the design doc for when we are comfortable with the state of the interface.

core/engine/engine.go
```diff
@@ -1,99 +1,166 @@
...
+	// Update the latest index
+	iLnk, err := indexLnk.AsLink()
+	if err != nil {
+		return err
+	}
+	err = e.putLatestIndex(iLnk.(cidlink.Link).Cid)
+	if err != nil {
+		return err
+	}
+	adv, err := schema.NewAdvertisement(e.privKey, latestAdvID, indexLnk, e.host.ID().String(), graphSupport)
```
masih, line 156: What would happen if `e.putLatestIndex` succeeds but we fail to advertise?

core/engine/engine_test.go
```diff
@@ -94,6 +94,10 @@ func genRandomIndexAndAdv(t *testing.T, e *Engine) (schema.Index, schema.Link_In
 	return index, indexLnk, adv, advLnk
 }
 
+/* NOTE: This test is flaky due to something deep in the task. Commenting it for now.
+Will work on fixing it and understand if it is something to do with our implementation
+or something deeper.
+
```
masih, line 100: Would it make sense to use `t.Skip` with an explanation of why it is skipped? See [docs](https://pkg.go.dev/testing#hdr-Skipping)

adlrocha: Good point, that's better. Thanks!

adlrocha:
@masih, for some reason that I don't understand the CI hates me, everything works locally but fails up. In any case, as this will change a lot feel free to merge it so we can build upon the interface.

gammazero:
> @masih, for some reason that I don't understand the CI hates me, everything works locally but fails up. In any case, as this will change a lot feel free to merge it so we can build upon the interface.

If you want to see the error locally, then run the same command as CI runs:
```sh
go test -v -coverprofile module-coverage.txt ./...
```

adlrocha:
Awesome, thanks @gammazero! In any case, I still don't get why is failing. The tests seem to pass fine. I'll look into it tomorrow.

gammazero:
Something is causing `go test -coverprofile=xx` to fail.  Maybe @mvdan has some idea.

adlrocha:
@masih, @gammazero, I really appreciate your help tracking the issue :) I think this is ready to merge.

command/daemon.go
```diff
@@ -68,8 +104,8 @@ func daemonCommand(cctx *cli.Context) error {
 
 	log.Infow("Shutting down daemon")
 
-	ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
-	defer cancel()
+	_, shCancel := context.WithTimeout(context.Background(), shutdownTimeout)
```
willscott, line 107: should we not get rid of this entirely if the derived context is no longer used?

willscott (approved):

command/daemon.go
adlrocha: I left a few placeholders there in case we needed them in  the future, but I agree is better to keep it clean and add it if we need it. Fixed.

willscott:
i suspect the interface for notifying on adding / removing car's / batches of CIDs will change again still, but this seems fine in terms of thinking about stubs for this iteration.

in particular, we need to understand what the right info will be for the go-data-transfer / graphsync request back to the provider. In the case of a miner, we conceptually need to know the `pieceID` for this, but more generally need to understand if we plan to do that by offering a `nil` go-data-transfer voucher and negotiating at the graphsync level, or if we intend to have a voucher pass through the metadata mechanism

willscott:
we will need some place in the metadata structure we use in this provider/client round-trip to potentially indicate the "price-per-byte" that retrieval would cost, for cases where it isn't provided for free. For this provider, it'll always be 0 / free, but we want to be writing a client that does graphsync provider fetches, so the logic on the client should be compatible with the integration of this to filecoin providers.

adlrocha:
> in particular, we need to understand what the right info will be for the go-data-transfer / graphsync request back to the provider. In the case of a miner, we conceptually need to know the pieceID for this, but more generally need to understand if we plan to do that by offering a nil go-data-transfer voucher and negotiating at the graphsync level, or if we intend to have a voucher pass through the metadata mechanism

This is for the actual retrieval of content and not the ingestion of the list of CIDs in a CAR by the indexer, right? i.e the information that needs to be added to metadata in order for the client to know how to make the retrieval after querying the indexer. So your point is that in the next iteration we'll need to figure out what metadata to include in `NotifyPut`. (just wanted to double-check if I understood it correctly)

masih:
> @masih, @gammazero, I really appreciate your help tracking the issue :) I think this is ready to merge.

@gammazero did all the work :)

PR #14, opened 2021-08-31
Implement libp2p server

adlrocha:
This PR:
- Implements libp2p server and client for indexer reference provider.
- Exposes `GetLatestAdv` and `GetAdv` through the server. These functions will be used by indexers to orchestrate provider sync.


api/v0/client/libp2p/client.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type Provider struct {
+	p2pc *libp2pclient.Client
+}
+
+func NewProvider(ctx context.Context, h host.Host, options ...libp2pclient.ClientOption) (*Provider, error) {
```
masih, line 20: Subjective comment: I wonder if having the word `Client` in the name of `Provider` client interface would read better from the perspective of developers using this library.

api/v0/common.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func DecodeError(data []byte) error {
+	if len(data) == 0 {
+		return nil
+	}
+	var e Error
+	err := json.Unmarshal(data, &e)
+	if err != nil {
+		return fmt.Errorf("cannot decode error message: %s", err)
```
masih, line 51: I am curious why `EncodeError` would return `serverError` but not here?

api/v0/common.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+type Error struct {
+	Message string
```
masih, line 12: Nonblocker: I would argue having a dedicated code field here would be a good investment, since `MakeError` already takes `status` int.

internal/libp2pclient/options.go
```diff
@@ -0,0 +1,32 @@
+package libp2pclient
+
+import (
+	"fmt"
+)
+
+// NOTE: Placeholder to add client options
```
masih, line 7:
```suggestion
// TODO: Placeholder to add client options
```

internal/libp2pclient/peer_message_sender.go
```diff
@@ -0,0 +1,141 @@
...
+	ms.r = msgio.NewVarintReaderSize(nstr, network.MessageSizeMax)
+	ms.stream = nstr
+
+	return nil
+}
+
+// sendMessage to a peer without waiting for any response.
+func (ms *peerMessageSender) sendMessage(ctx context.Context, msg proto.Message, h host.Host, protos []protocol.ID) error {
+	err := ms.ctxLock.Lock(ctx)
+	if err != nil {
```
masih, line 76:
```suggestion
	if err := ms.ctxLock.Lock(ctx); err != nil {
```

internal/libp2pclient/peer_message_sender.go
```diff
@@ -0,0 +1,141 @@
...
+		var data []byte
+		data, err = r.ReadMsg()
+		defer r.ReleaseMsg(data)
+		if err != nil {
+			return
+		}
+		err = decodeRsp(data)
+	}(ms.r)
+
+	t := time.NewTimer(readMessageTimeout)
```
masih, line 129: Why a dedicated read timeout? Could a user not achieve this with `context.WithTimeout`? 

If the purpose is to have a maximum read timeout limit, then would it make sense to rename `readMessageTimeout` to something that makes it clear it's the _maximum_ read timeout?

internal/libp2pserver/server.go
```diff
@@ -0,0 +1,94 @@
...
+		_ = stream.Close()
+	} else {
+		// otherwise, send an error.
+		_ = stream.Reset()
+	}
+}
+
+// Returns true on orderly completion of writes (so we can Close the stream conveniently).
+func (s *Server) handleNewMessages(stream network.Stream) bool {
+	ctx := s.ctx
```
masih, line 65: Is this a common practice? I am curious why we don't take context as an argument.

internal/libp2pserver/server.go
```diff
@@ -0,0 +1,94 @@
...
+		resp, err := handler.HandleMessage(ctx, mPeer, msgbytes)
+		r.ReleaseMsg(msgbytes)
+		if err != nil {
+			return true
+		}
+
+		// send out response msg
+		err = p2putil.WriteMsg(stream, resp)
+		if err != nil {
+			return false
```
masih, line 91: Log error?

internal/libp2pserver/server.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			r.ReleaseMsg(msgbytes)
+			return err == io.EOF
+		}
+		timer.Reset(streamIdleTimeout)
+
+		resp, err := handler.HandleMessage(ctx, mPeer, msgbytes)
+		r.ReleaseMsg(msgbytes)
+		if err != nil {
+			return true
```
masih, line 85: Log error?

masih (commented):
On potential blocker on the `proto` message schema; left a question. Otherwise all suggestions 🍻

api/v0/pb/provider.proto
```diff
@@ -0,0 +1,23 @@
+// In order to re-generate the golang packages for `Message` you will need...
+// 1. Protobuf binary (tested with protoc 3.0.0). - https://github.com/gogo/protobuf/releases
+// 2. Gogo Protobuf (tested with gogo 0.3). - https://github.com/gogo/protobuf
+// Now from `libp2p/<path>/pb` you can run...
+// `protoc --gogo_out=. --proto_path=../../<path>/pb/ --proto_path=./ --proto_path=/usr/include pb.proto`
+
+syntax = "proto3";
+package reqresp.pb;
```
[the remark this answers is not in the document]
adlrocha: Actually we don't need the package in the protobuf definition (I don't expect it to be used as part of another protobuf). Removing it.

api/v0/pb/Makefile
```diff
@@ -0,0 +1,11 @@
...
+GO = $(PB:.proto=.pb.go)
+
+all: $(GO)
+
+%.pb.go: %.proto
+	protoc --proto_path=$(GOPATH)/pkg/mod:. --proto_path=/usr/include --gogofaster_out=. $<
+
+clean:
+	rm -f *.pb.go
+	rm -f *.go
```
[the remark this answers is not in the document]
adlrocha: Wow, that shouldn't be there. Thanks!

api/v0/pb/provider.proto
```diff
@@ -0,0 +1,23 @@
...
+
+message ProviderMessage {
+    enum MessageType {
+        ERROR_RESPONSE = 0;
```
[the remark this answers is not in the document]
adlrocha: > This means if unset, the value of MessageType will be ERROR_RESPONSE; i.e. we can't differentiate on the receiving end whether it was unset or explicitly set to ERROR_RESPONSE. Is that intentional?

We do, responses are defaulted to an `ERROR_RESPONSE` with `"unsupported message type"`. But happy to add `UNSET = 0` as a sanity-check.

api/v0/models/models.go
```diff
@@ -0,0 +1,67 @@
+package models
```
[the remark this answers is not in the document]
adlrocha: We decided to include in `api/` every public asset clients or other implementations of the protocol would need. This is why these models along with the client and the pb are included here. We are following the same approach in `storetheindex`. No strong opinions, so happy to open the discussion again.

internal/libp2pclient/peer_message_sender.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (ms *peerMessageSender) ctxReadMsg(ctx context.Context, decodeRsp DecodeResponseFunc) error {
+	done := make(chan struct{})
+	var err error
+	go func(r msgio.ReadCloser) {
+		defer close(done)
+		var data []byte
+		data, err = r.ReadMsg()
+		defer r.ReleaseMsg(data)
```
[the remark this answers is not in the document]
adlrocha: Aren't we releasing it at the end by deferring it?  (I don't think it would change much though)

api/v0/client/libp2p/client.go
adlrocha: You're gonna laugh, but I stared at this definition wondering the same for a few minutes. I settled for `NewProvider` to make it shorter. But if you are also wondering this, it means it has to be `NewProviderClient`. Thanks!

api/v0/common.go
adlrocha: @gammazero and I had this same discussion yesterday when you left from the call. At this point we don't use codes for anything. We most probably add it in the future.

internal/libp2pclient/peer_message_sender.go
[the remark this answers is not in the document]
masih: Right. The question on my part was: would there be anything _to release_ if error is non-nil? Please feel free to ignore this comment if `ReleaseMsg` won't error if data is empty etc.

api/v0/models/models.go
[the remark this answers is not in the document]
masih: nope all good thank you for explaining that 👍

api/v0/pb/provider.proto
```diff
@@ -0,0 +1,23 @@
...
+// 1. Protobuf binary (tested with protoc 3.0.0). - https://github.com/gogo/protobuf/releases
+// 2. Gogo Protobuf (tested with gogo 0.3). - https://github.com/gogo/protobuf
+// Now from `libp2p/<path>/pb` you can run...
+// `protoc --gogo_out=. --proto_path=../../<path>/pb/ --proto_path=./ --proto_path=/usr/include pb.proto`
+
+syntax = "proto3";
+package reqresp.pb;
+
+message ProviderMessage {
+    enum MessageType {
```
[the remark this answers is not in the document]
adlrocha: Flexibility? Ease of implementation? I just keep seeing this approach for libp2p protocols and I have internalized it. See for example https://github.com/libp2p/go-libp2p-kad-dht/blob/master/pb/dht.proto.

internal/libp2pserver/server.go
adlrocha: Unfortunately it is. Libp2p handlers can only receive a stream as input in its signature.

api/v0/pb/provider.proto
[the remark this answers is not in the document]
masih: OK thanks. this seemed slightly odd to me, since we pay the cost of depending on protobuf but don't take full advantage of what it can offer: backward compatibility, contract-first design, message schema conformity etc.

api/v0/common.go
adlrocha: `EncodeErorr` is used to send the error through the wire. `DecodeError` is used internally.

internal/libp2pclient/client.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+var log = logging.Logger("libp2pclient")
+
+// Client Libp2pclient is responsible for sending
+// requests to other peers.
+type Client struct {
+	ctx  context.Context
```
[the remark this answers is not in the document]
adlrocha: Good catch. It is not used at all.

adlrocha:
Thank you for such a thorough review, @masih. Let me know if there is anything else before merging.

masih (approved):

PR #19, opened 2021-09-08
Admin server, connect command in CLI

adlrocha:
This PR:
- Implements an http server for admin purposes.
- Implements a command to attempt connections to indexer nodes (or any other libp2p node)
- Does some housekeeping to the daemon.

### DIFF
--- a/command/daemon.go
+++ b/command/daemon.go
@@ -5,13 +5,17 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/filecoin-project/indexer-reference-provider/config"
 	"github.com/filecoin-project/indexer-reference-provider/core/engine"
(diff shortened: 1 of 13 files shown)


core/engine/engine_test.go
```diff
@@ -67,8 +67,8 @@ func mkTestHost() host.Host {
 // testing purposes. A more complex callback could read
 // feom the CID index and return the list of CIDs.
 func simpleCb(cids []cid.Cid) core.CidCallback {
-	return func(c cid.Cid) []cid.Cid {
-		return cids
+	return func(c cid.Cid) ([]cid.Cid, error) {
```
willscott: eventually we'll want the `[]cid.Cid` to be instead perhaps a `chan cid.Cid` so we can stream the list as we need it?

willscott (approved):

PR #21, opened 2021-09-09
Add lookupKey. Key-Cid mapping

adlrocha:
This PR:
- Adds a `lookupKey` to identify the list of CIDs being advertised.
- Handles the relationship between `lookupKey` and list of CIDs, and generates the link for the list of CIDs structure so it can be ingested by indexers.

### DIFF
--- a/core/engine/engine.go
+++ b/core/engine/engine.go
@@ -1,9 +1,11 @@
 package engine
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"errors"
+	"io"
 	"sync"
 
 	"github.com/filecoin-project/indexer-reference-provider/config"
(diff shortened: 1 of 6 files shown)


core/engine/engine.go
```diff
@@ -197,10 +201,79 @@ func (e *Engine) GetLatestAdv(ctx context.Context) (cid.Cid, schema.Advertisemen
...
+func noStoreLinkSystem() ipld.LinkSystem {
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.StorageWriteOpener = func(lctx ipld.LinkContext) (io.Writer, ipld.BlockWriteCommitter, error) {
+		buf := bytes.NewBuffer(nil)
+		return buf, func(lnk ipld.Link) error {
+			return nil
+		}, nil
+	}
+	return lsys
+}
```
willscott, line 215: @warpfork is this something that should be a default stub in ipld-prime?

adlrocha: I actually thought that `cidlink.DefaultSystem()` did the trick if now `StorageWriterOpener` was set. Maybe we can follow this approach.

masih (approved):
Approving to keep things going since other parts need updating due to interface changes.

PR #24, opened 2021-09-09
Implement `import car` CLI command

masih:
Implement `import car` command that takes a path and optional lookup
key and metadata, then connects to the admin HTTP API to import it.

### DIFF
--- a/command/import.go
+++ b/command/import.go
@@ -1,0 +1,98 @@
+package command
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	adminserver "github.com/filecoin-project/indexer-reference-provider/server/http"
+	"github.com/urfave/cli/v2"
+)
+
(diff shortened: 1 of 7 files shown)


command/import.go
```diff
@@ -0,0 +1,98 @@
...
+
+	httpReq.Header.Set("Content-Type", "application/json")
+	cl := &http.Client{}
+	resp, err := cl.Do(httpReq)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to import car, due to failure response from Admin server: %v", http.StatusText(resp.StatusCode))
```
willscott, line 78: will there only ever be an error code and no body of the response with the specific error that you want to relay?

willscott (approved):

command/import.go
masih: On the admin server side there is no error message structure yet. happy to pick that up after this PR is merged to convey more meaningful messages about erroneous cases.

adlrocha (approved):

PR #25, opened 2021-09-10
Containerize the provider by defining a `Dockerfile`

masih:
Add a dockerfile that builds the reference provider with a conditional
initialiazation build argument that is enabled by default. Note, the
docker image built as a result is meant for testing purposes only, since
once initialized it will contain a generated private key that is unsafe
to use in production.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM golang:1.16-buster as build
+
+WORKDIR /go/src/provider
+
+COPY go.mod go.sum ./
+RUN go get -d -v ./...
+
+ADD . /go/src/provider
+RUN go build -o /go/bin/provider
+
+# TODO consider auto initialization flag as part of `daemon` command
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN go get -d -v ./...
+
+ADD . /go/src/provider
+RUN go build -o /go/bin/provider
+
+# TODO consider auto initialization flag as part of `daemon` command
+ARG INIT_PROVIDER='true'
+RUN if test "${INIT_PROVIDER}" = 'true'; then /go/bin/provider init; else echo 'skipping provider initialization.'; fi
+
+FROM gcr.io/distroless/base-debian11:debug
```
willscott, line 15: 'debug' as the production image?
we may need ca-certs in the image to validate the libp2p connection security?

willscott (approved):

PR #31, opened 2021-09-12
Refactor admin server config and `server` package structure

masih:
Introduce a dedicated config struct that captures the configurable
values of the admin server.

Move server and client packages to `provider` subdirectory to follow
consistent convention.


core/engine/engine_test.go
```diff
@@ -165,6 +165,7 @@ func TestPublishLocal(t *testing.T) {
 }
 
 func TestNotifyPublish(t *testing.T) {
+	skipFlaky(t)
```
adlrocha, line 168: I think all tests using Gossipsub will end up being flaky as long as we don't find a way to make them async. I'll check how is this being tested in the gossipsub basecode (at this pace we'll end up skipping all tests) :)

adlrocha (approved):
LGTM

PR #39, opened 2021-09-17
core/engine: ignore fewer errors

mvdan:
(see commit message)

### DIFF
--- a/core/engine/engine_test.go
+++ b/core/engine/engine_test.go
@@ -59,8 +59,9 @@ func mkMockSubscriber(t *testing.T, h host.Host) (legs.LegSubscriber, *legs.LegT
 	return ls, lt
 }
 
-func mkTestHost() host.Host {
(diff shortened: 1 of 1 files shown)


core/engine/engine_test.go
```diff
@@ -60,7 +60,10 @@ func mkMockSubscriber(t *testing.T, h host.Host) (legs.LegSubscriber, *legs.LegT
 }
 
 func mkTestHost() host.Host {
```
masih, line 62: Would it make sense to take `testing.T` here and assert no error through it rather than panicking? 

Ditto for other functions used only in tests.

masih (approved):

core/engine/engine_test.go
mvdan: I thought about that but was being lazy :) will do

masih (approved):

PR #47, opened 2021-10-02
Use existing data transfer

hannahhoward:
# Goals

move building of graphsync and data transfer out of the engine itself, since these will need to be
dependencies in the eventual lotus integration

# Implementation

- Update go-legs
- Update go-data-transfer & go-graphsync to Lotus level versions
- Remove leg transport (no longer exists)
- add DataTransfer as parameter to engine setup
- setup standalone Data transfer manually where needed


command/daemon.go
```diff
@@ -92,8 +100,22 @@ func daemonCommand(cctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
+	gsnet := gsnet.NewFromLibp2pHost(h)
+	gs := gsimpl.New(context.Background(), gsnet, cidlink.DefaultLinkSystem())
+	tp := gstransport.NewTransport(h.ID(), gs)
+	dtNet := dtnetwork.NewFromLibp2pHost(h)
+	tmpDir, err := ioutil.TempDir("", "indexer-dt-dir")
```
gammazero, line 108: Is this a directory that we may want to reuse across separate runs of the reference-provider daemon?

gammazero (approved):
Looks good.  Have one question.

command/daemon.go
hannahhoward: eventually yes, but it will also be dependent on who's running it. My understanding is that in Lotus, we won't actually be running the Daemon cause it will just be part of the markets binary.

PR #48, opened 2021-10-04
Refactor code structure and add documentations

masih:
Refactor the interface to follow the same naming convention as other
indexer related repos, i.e. `provider.Interface` instead of
`core.Interface`.

Improve documentation on the interface and leave TODO on outstanding
items. 

Remove unused `PushAdv` from the interface. Remove `Push` from the 
interface since provider will no longer need to satisfy that usecase.

Move CLI related commands to `cmd` folder to follow conventions in other
reopositiories.

Regenerate mock interface and reflect changes across the tests.


interface.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Close closes this provider, discarding all resources occupied by it.
+	// After calling this function the provider is no longer available for use.
+	Close(ctx context.Context) error
+}
+
+// Callback is used by provider to look up a list of multihashes associated to a key.
+// The callback must produce the same list of multihashes for the same key.
+// See: Interface.NotifyPut, Interface.NotifyRemove
+// TODO: update docs once the iterator return type refactor is done.
```
masih, line 75: To avoid extra noise here and to explicitly document the change in a commit of its own I will work on this in a separate PR.

interface.go
```diff
@@ -0,0 +1,74 @@
...
+	// Therefore, a Callback must be registered prior to using this function.
+	//
+	// The metadata is serialized data that provides hits about how to retrieve data and is entirely protocol dependant.
+	// The metadata may be nil and is optional.
+	// If specified, the metadata value must start with a uvarint that correspons to the indexer.Value protocol ID.
+	//
+	// This function returns the ID of the advertisement published.
+	//
+	// See: https://pkg.go.dev/github.com/filecoin-project/go-indexer-core
+	// TODO: update pkg link above once encodeMetadata is exported and docs are updated.
```
masih, line 49: @gammazero  Ieft a TODO here for myself to update the link once encodeMetadata is exported. Please shout if you can think of a better URL to put in there just now. Cheers

engine/engine.go
```diff
@@ -5,44 +5,33 @@ import (
...
-	// MaxCidsInChunk number of entries to include in each chunk
-	// of ingestion linked list.
-	MaxCidsInChunk = 100
-)
 
-var (
-	// ErrNoCallback is thrown when no callback has been defined.
-	ErrNoCallback = errors.New("no callback was registered in indexer")
-)
+// MaxCidsInChunk number of entries to include in each chunk
```
mvdan, line 29:
```suggestion
// MaxCidsInChunk is the maximum number of entries to include in each chunk
```

engine/engine.go
```diff
@@ -143,11 +132,6 @@ func (e *Engine) PublishLocal(ctx context.Context, adv schema.Advertisement) (ci
 	return c, nil
 }
 
-// Publish publishes the advertisement on the pubsub channel.
-//
-// Advertisements are published immutably, so it doesn't matter if the same
-// advertisement is published twice, as this will only propagate the
-// announcement again through the pubsub channel.
```
mvdan, line 150: hm, removed godoc on purpose?

mvdan: Ah, right, it's documented in the interface - SGTM.

interface.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// GetAdv gets the advertisement that corresponds to the given id.
+	GetAdv(ctx context.Context, id cid.Cid) (schema.Advertisement, error)
+
+	// GetLatestAdv gets the latest advertisement on this provider's advertisement chain and the CID to which it corresponds.
+	GetLatestAdv(ctx context.Context) (cid.Cid, schema.Advertisement, error)
+
+	// Close closes this provider, discarding all resources occupied by it.
+	// After calling this function the provider is no longer available for use.
+	Close(ctx context.Context) error
```
mvdan, line 67: Since we're here, I think we should rename this to Shutdown, given that it takes a context and presumably blocks until all goroutines have stopped. A Close operation tends to be immediate and not take a context.

mvdan (approved):

interface.go
```diff
@@ -0,0 +1,74 @@
...
+	// RegisterCallback registers the callback used by the provider to look up a list of multihashes by LookupKey.
+	// Only a single callback is supported; repeated calls to this function will replace the previous callback.
+	RegisterCallback(cb Callback)
+
+	// NotifyPut sginals to the provider that the list of multihashes looked up by the given key are available.
+	// The given key is then used to look up the list of multihashes via Callback.
+	// An advertisement is then generated, appended to the chain of advertisements and published onto the gossip pubsub channel.
+	// Therefore, a Callback must be registered prior to using this function.
+	//
+	// The metadata is serialized data that provides hits about how to retrieve data and is entirely protocol dependant.
```
mvdan, line 42: "hits" is a typo for "hints" maybe? :)

interface.go
```diff
@@ -0,0 +1,74 @@
...
+	// The metadata may be nil and is optional.
+	// If specified, the metadata value must start with a uvarint that correspons to the indexer.Value protocol ID.
```
mvdan, line 44: I wonder if taking `metadataProtocol uint64, metadataBody []byte` instead of a joint `[]byte` would be less prone to misuse.

Alternatively, we could document that implementations of NotifyPut should error if the metadata doesn't start with a valid uvarint.

masih: Yeah I think it might be worth separating those.

> Alternatively, we could document that implementations of NotifyPut should error if the metadata doesn't start with a valid uvarint.

Not sure if we can correctly assert this; a serialised metadata without protocol code _could_ still start with variant. 

Going with separation if you agree? I see the same pattern in [indexer-core](https://github.com/filecoin-project/go-indexer-core/blob/main/value.go#L20).

mvdan: Separation SGTM. I agree the sanity check would have false negatives, but with a good enough rate of true positives, I imagine it could still catch some bugs.

masih: On second thought I think we should define a dedicated type for metadata in go-indexer-core that has receivers for marshalling/unmarshalling. This is a type used in both indexer node and provider and seems like it belongs to core.

Functionality is already implemented as part of [`indexer.Value`](https://github.com/filecoin-project/go-indexer-core/blob/main/value.go#L46-L53) It just needs to be refactored into its own type. 

@gammazero  what are your thoughts? I am happy to pick this up

masih: Captured in a separate issue and left TODO in code: 
* https://github.com/filecoin-project/go-indexer-core/issues/26

engine/engine.go
masih: I ended up unexporting this and dropped `Cid` in name.

PR #50, opened 2021-10-07
Add contextID to advertisements and indexer ingest requests

gammazero:
This PR adds a context ID to advertisements and indexer ingest requests.  The context ID allows the bulk update or removal of metadata for a provider

Note [here](https://github.com/filecoin-project/indexer-reference-provider/pull/50/files#diff-2e6c4934f63ff9b712c2c346b33036af4724adf70b0801fff9b74f71b37fcd89R302) that the advertisement lookup key is used as the context ID.  Need reviewers to evaluate if this is the correct thing to do here, or if a separate key is needed. @adlrocha Is this what you had originally intended?
 
This PR also changes the reference provider to use new indexer find result that does not separate provider address info.

This PR should remain draft until tests are in place to test context ID functionality.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -185,7 +185,6 @@ github.com/filecoin-project/go-ds-versioning v0.1.0 h1:y/X6UksYTsK8TLCI7rttCKEvl
 github.com/filecoin-project/go-ds-versioning v0.1.0/go.mod h1:mp16rb4i2QPmxBnmanUx8i/XANp+PFCCJWiAb+VW4/s=
 github.com/filecoin-project/go-indexer-core v0.2.0 h1:I3mtrTLH6aXOkbxsjRKv3rn8tLgoa6vIYl8XuBNRDik=
 github.com/filecoin-project/go-indexer-core v0.2.0/go.mod h1:CsD+HUiCB+lwL+mxXPKQCsqoA9fezXTheU7IjsFohCU=
-github.com/filecoin-project/go-legs v0.0.0-20210922204025-c6f68b62ab16 h1:xFiCa75M7kjr9/AfaVvFNjJKsMgtIP25g+DXKx4Ub7A=
 github.com/filecoin-project/go-legs v0.0.0-20210922204025-c6f68b62ab16/go.mod h1:qLH/nW+bQoOmnLG4eGU36vubv0MmAezQ2Y2FQBa1hM8=
 github.com/filecoin-project/go-legs v0.0.0-20211002011713-f1f52b88c467 h1:zJYJcnbujeGTfLtr9ErJa0+oFBtY6mqW+7dCLf4/pQU=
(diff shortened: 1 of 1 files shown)


adlrocha:
> Note here that the advertisement lookup key is used as the context ID. Need reviewers to evaluate if this is the correct thing to do here, or if a separate key is needed. @adlrocha Is this what you had originally intended?

Correct. That was the idea to use the contextID as the lookupKey to make it available for the indexer. I've seen that you've included already in storetheindex's schema the `contextID` so it can be ingested in the indexer.

I see that the PR is still a draft, let me know if you want me to approve it (but everything LGTM).

engine/engine.go
```diff
@@ -284,8 +299,9 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, key provider.LookupKey,
 		previousLnk = nb.Build().(schema.Link_Advertisement)
 	}
 
+	contextID := []byte(key)
```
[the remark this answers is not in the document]
masih: I am off on this one. Let's get rid of the lookup key completely and call it what it is `ContextID`. I am happy to do that in a separate PR. This PR looks fine, thanks @gammazero 🍻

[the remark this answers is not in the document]
adlrocha: I agree with @masih. We can definitely get rid of the LookuKey at this point and call it `contextID`. No need for the lookupKey anymore.

go.mod
```diff
@@ -4,9 +4,9 @@ go 1.16
 
 require (
 	github.com/filecoin-project/go-data-transfer v1.10.1
-	github.com/filecoin-project/go-indexer-core v0.1.5
+	github.com/filecoin-project/go-indexer-core v0.2.0
 	github.com/filecoin-project/go-legs v0.0.0-20211002011713-f1f52b88c467
-	github.com/filecoin-project/storetheindex v0.0.0-20211001170917-11797057b28d
+	github.com/filecoin-project/storetheindex v0.0.0-20211008013136-08d087cd7027
```
willscott, line 9: ideally we don't need to depend on the 'storetheindex' as well as 'go-indexer-core' we should understand where the dep came from.

gammazero: The dependency on go-indexer-core is for data type definitions, such as `indexer.Value` and `indexer.Metadata`.  The dependency on storetheindex is due to interface and API definitions.

It should be possible to only depend on storetheindex by removing the use `indexer.Value` (which should only be used between storetheindex and go-indexer-core), and by moving the definition of `indexer.Metadata` into storetheindex.  The second part doable since go-indexer-core only cares about the `indexer.Value` and does not need to see the metadata structure.

gammazero: @willscott This is address by https://github.com/filecoin-project/go-indexer-core/pull/30 and  https://github.com/filecoin-project/storetheindex/pull/93, and will be addressed by a subsequent PR in this repo.

PR #51, opened 2021-10-11
Use iterator instead of double channel in `provider.Callback`

masih:
Introduce `MultihashIterator` type to replace the double channel
returned by `provider.Callback`. This is to simplify the usage for
clients.

Require callback to take context in order to gracefully handle the
lifecycle of gorutines that may be started as part of
`MultihashIterator`.

Implement utility class that converts CARv2 `MultihashSortedIndex` into
a `MultihashIterator`. This type is not exported. But may be in the
future if useful to existing clients.

Refactor use of "cid" in names now that provider works with multihashes.

### DIFF
--- a/engine/linksystem.go
+++ b/engine/linksystem.go
@@ -106,12 +106,15 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
 			// Entries link, if just gets the cid, and uses its local map cid
 			// to lookupKey to trigger the removal of all entries for that
 			// lookupKey in its index.
-			chcids, cherr := e.cb(key)
+			mhIter, err := e.cb(lctx.Ctx, key)
+			if err != nil {
+				return nil, err
+			}
 
 			// Store the linked list entries in cache as we generate them.  We
 			// use the cache linksystem that stores entries in an in-memory
 			// datastore.
-			_, err = generateChunks(e.cachelsys, chcids, cherr, maxIngestChunk)
+			_, err = generateChunks(e.cachelsys, mhIter, maxIngestChunk)
 			if err != nil {
 				log.Errorf("Error generating linked list from callback: %s", err)
 				return nil, err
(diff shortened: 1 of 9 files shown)


engine/linksystem.go
```diff
@@ -151,56 +154,39 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
...
-				break eachCid
-			}
-			mhs = append(mhs, ec)
-		case err = <-cherr:
-			if err != nil {
-				// If something in error channel return error
-				return nil, err
+		next, err := mhIter.Next()
+		if err != nil {
+			if err == io.EOF {
```
mvdan, line 164: what if you make the EOF case a `break`, then you have the "we've stopped" logic further down? otherwise the `len(mhs) != 0` logic is very indented

engine/linksystem.go
```diff
@@ -151,56 +154,39 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
...
 		}
+		mhs = append(mhs, next)
+		totalMhCount++
 
-		// Create chunk of cids and link to previous chunk
-		if len(mhs) == numEntries {
-			cidCount += len(mhs)
-			chunkCount++
-			// Create the chunk and restart variables
+		if len(mhs) == maxChunkSize {
```
mvdan, line 179: nit: I know this code is older, but it would be nicer to write it defensively with a `>=`, in case our code is buggy and we increment by more than one or partially skip an iteration.

interface.go
```diff
@@ -68,8 +68,15 @@ type Interface interface {
 	Shutdown(ctx context.Context) error
 }
 
+// MultihashIterator iterates over a list of multihashes.
+type MultihashIterator interface {
+	// Next returns the next multihash in the list of mulitihashes.
+	// This function return io.EOF when there are no more multihashes to return.
```
mvdan, line 74: I'd say "a zero multihash and io.EOF" to clarify that EOF is never returned along with the last value.

See io.Reader, which *can* give you the last chunk of bytes along with EOF to save a call.

internal/suppliers/index_mh_iter.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func (i indexMhIterator) Next() (multihash.Multihash, error) {
+	if next, ok := <-i.mhch; ok {
+		return next, nil
+	}
+	if err, ok := <-i.errch; ok {
+		return nil, err
+	}
```
mvdan, line 48: I'd argue you need to check the error channel before the multihash channel. If we've encountered an error, such as a timeout, we want to report that immediately - instead of returning the last buffered multihash, and then the error. Probably document that in the code, too.

mvdan (approved):
Just nits :) LGTM

mvdan (approved):
SGTM. Up to you if you want to keep the tightly coupled unit tests :)

internal/suppliers/index_mh_iter_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.Equal(t, io.EOF, err)
+}
+
+func TestIndexMhIterator_NextReturnsErrorOnUnHappyPath(t *testing.T) {
+	wantMh, err := multihash.Sum([]byte("fish"), multihash.SHA3_256, -1)
+	require.NoError(t, err)
+	wantErr := errors.New("lobster")
+
+	subject := newIndexMhIterator(context.Background(), &testIterableIndex{
+		MultihashIndexSorted: nil,
```
mvdan, line 50: FYI, you can drop these lines, as the zero value is already nil

mvdan (approved):
:ok_hand: very nicely done with the func field, I hadn't thought of that

PR #54, opened 2021-10-13
Replace `LookupKey` concept with `ContextID`

masih:
Remove the `LookupKey` type in favour of `[]byte` that represents
"Context ID", an ID used already bo go-indexer-core to aggregate
metadata. For provider's purposes this results in 1:1 functionaly with
`LookupKey`, since a "Context ID" will always return the same list of
multihashes.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -147,16 +147,16 @@ func (e *Engine) RegisterCallback(cb provider.Callback) {
 	e.cb = cb
 }
 
-func (e *Engine) NotifyPut(ctx context.Context, key provider.LookupKey, metadata stiapi.Metadata) (cid.Cid, error) {
-	log.Debugf("NotifyPut for lookup key")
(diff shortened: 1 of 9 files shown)


engine/engine.go
```diff
@@ -147,16 +147,16 @@ func (e *Engine) RegisterCallback(cb provider.Callback) {
 	e.cb = cb
 }
 
-func (e *Engine) NotifyPut(ctx context.Context, key provider.LookupKey, metadata stiapi.Metadata) (cid.Cid, error) {
-	log.Debugf("NotifyPut for lookup key")
+func (e *Engine) NotifyPut(ctx context.Context, contextId []byte, metadata stiapi.Metadata) (cid.Cid, error) {
```
mvdan, line 150: nit: I think most people will expect the spelling to be contextID, just like we type DoHTTP instead of DoHttp.

mvdan (approved):

PR #59, opened 2021-10-14
Export utility to construct`MultihashIterator` from CAR `IterableIndex`

masih:
Provide an implementation of `MultihashIterator` from CARv2 index. This
is useful to the clients since most will be dealing with CAR
files as a way to supply multihashes to the provider.

### DIFF
--- a/index_mh_iter.go
+++ b/index_mh_iter.go
@@ -1,22 +1,27 @@
-package suppliers
+package provider
 
 import (
 	"context"
 	"io"
 
-	provider "github.com/filecoin-project/indexer-reference-provider"
-	"github.com/ipld/go-car/v2/index"
+	carindex "github.com/ipld/go-car/v2/index"
 	"github.com/multiformats/go-multihash"
 )
 
-var _ provider.MultihashIterator = (*indexMhIterator)(nil)
+var _ MultihashIterator = (*indexMhIterator)(nil)
 
 type indexMhIterator struct {
 	mhch chan multihash.Multihash
 	err  error
 }
 
-func newIndexMhIterator(ctx context.Context, idx index.IterableIndex) *indexMhIterator {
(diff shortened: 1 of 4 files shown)


index_mh_iter.go
```diff
@@ -1,22 +1,22 @@
...
-var _ provider.MultihashIterator = (*indexMhIterator)(nil)
+var _ MultihashIterator = (*indexMhIterator)(nil)
 
 type indexMhIterator struct {
 	mhch chan multihash.Multihash
 	err  error
 }
 
-func newIndexMhIterator(ctx context.Context, idx index.IterableIndex) *indexMhIterator {
+// NewMultihashIteratorFromIterableIndex constructs a new MultihashIterator from a CAR index.IterableIndex.
```
mvdan, line 18: This should document what the context is used for.

index_mh_iter.go
```diff
@@ -1,22 +1,22 @@
...
+func NewMultihashIteratorFromIterableIndex(ctx context.Context, idx index.IterableIndex) MultihashIterator {
```
mvdan, line 19: Hm, I find this name pretty verbose. How about `IterableIndexIterator`, `CarIndexIterator`, or `CarMultihashIterator`?

I'd also prefer to name the index import `carindex`, to clarify that the parameter is `carindex.IterableIndex`. Otherwise the name "index" is sort of confusing, as it doesn't clarify it means a CAR index.

mvdan: I think I prefer CarMultihashIterator, because it clarifies the input is CAR-specific, and it's still clear that the output is a MultihashIterator.

index_mh_iter.go
masih: Done; I'd love your quick eye on the docs before merging please 🍻

PR #60, opened 2021-10-15
Export explicit error types for unknwon `contextID`

masih:
Export an explicit error type to signal that a `contextID` is never seen
before if given to `NotifyRemove`.

While at it, move the known error type for no registered callback out of
engine and onto `provider` package.

Update docs and add tests.


mvdan (approved):
Want to ask Dirk's review too, since they'll use this?

PR #62, opened 2021-10-16
ensure CarMultihashIterator sorts CIDs by offset

mvdan:
(see commit message)

### DIFF
--- a/index_mh_iter.go
+++ b/index_mh_iter.go
@@ -2,7 +2,9 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"sort"
 
 	carindex "github.com/ipld/go-car/v2/index"
 	"github.com/multiformats/go-multihash"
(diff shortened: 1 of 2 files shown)


mvdan:
FYI @masih, Andrew correctly spotted a source of non-determinism when building IPLD nodes. Details in the commit message. This completely butchers our API decisions with the context etc, see the TODOs :) We can talk about it on Monday. This is a quick fix to unblock Andrew in the meantime.

gammazero (approved):
LGTM. This is probably more than a quick fix as it should ultimately be the provider's responsibility to ensure deterministic CAR iteration.

mvdan:
I agree that this approach is likely what we want anyway, long-term. The "quick fix" part is how the current API for loading the index no longer makes much sense :) I've left TODOs around that. Masih and I can discuss it tomorrow, and probably simplify it somewhat.

I'll merge for now, to just make the main branch work reliably sooner than later.

masih:
I see so the point is global determinism across all indices vs local determinism among indices of same type.

PR #66, opened 2021-10-20
Serve Car Files via Filecoin Retrieval Protocol

hannahhoward:
# Goals

Setup the provider so it can serve car files to retrieval clients

# Implementation

- Implement cardatatransfer, a module that serves files via Filecoin retrieval protocol (always free) on top of data transfer & graphsync. 
    - copy a lot of types over from markets
    - had to bring in cbor-gen unfortunately, bindnode is not quite ready to work with markets types. cc @mvdan for interesting challenges -- the biggest missing components I could see immediately is the lack of support for nillable struct values via pointer types, and the lack of an Any type in the current schema type list.
    - cardatatransfer is backed by a blockstore supplier, which is the CarSupplier
- Refactor CarSupplier to remove dual put methods, reverse lookup based on path, and simply use contextID for all methods (the backing reason is we really need to do the hashing of filepaths for implicit keys outside of this module, as we need them to generate metadata for Filecoin)
- import handler now just always assumes a key
- import command now autogenerates a key
- if no metadata is present, takes the specified or generated key and makes Filecoin V1 metadata for it so you can use with car supplier
- added a testutil directory to start sharing commonly used functions

# For discussion

- the command syntax for import is very much a guess as to what we want -- definitely open to feedback and revision
- you can largely gloss over the retrieval_types.go and the stores implementation -- these are copy pasted from markets
- I did not add a retrieve CLI command cause it seems odd to spin up a new libp2p host just to sub in for the future retrieval client. but we could.

### DIFF
--- a/internal/cardatatransfer/cardatatransfer_test.go
+++ b/internal/cardatatransfer/cardatatransfer_test.go
@@ -173,6 +173,11 @@ func TestCarDataTransfer(t *testing.T) {
 			dstDt := testutil.SetupDataTransferOnHost(t, dstHost, dstStore, lsys)
 			mn.LinkAll()
 
+			var expectedLen int
+			// read blockstore length ahead of time
+			if data.expectedBlockstoreResult != nil {
+				expectedLen = testutil.GetBstoreLen(ctx, t, data.expectedBlockstoreResult)
+			}
 			dstResultChan := make(chan bool, 1)
 			var dstMessage string
 			dstDt.SubscribeToEvents(func(event datatransfer.Event, channelState datatransfer.ChannelState) {
@@ -200,7 +205,6 @@ func TestCarDataTransfer(t *testing.T) {
 			case dstResult := <-dstResultChan:
 				require.Equal(t, data.expectSuccess, dstResult)
 				if data.expectSuccess {
-					expectedLen := testutil.GetBstoreLen(ctx, t, data.expectedBlockstoreResult)
 					receivedLen := testutil.GetBstoreLen(ctx, t, dstBlockstore)
 					require.Equal(t, expectedLen, receivedLen)
 				} else {


e2e_retrieve_test.go
```diff
@@ -0,0 +1,128 @@
...
+	s := p2pserver.New(ctx, h, e)
+
+	return s, h, cs, e
+}
+
+func setupClient(ctx context.Context, p peer.ID, t *testing.T) (datatransfer.Manager, *p2pclient.Client) {
+	store := dssync.MutexWrap(datastore.NewMapDatastore())
+	blockStore := blockstore.NewBlockstore(store)
+	lsys := storeutil.LinkSystemForBlockstore(blockStore)
+	h, err := libp2p.New(context.Background(), libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/1"))
```
willscott: should this be `/0`?

willscott (approved):
🚀

hannahhoward:
@rvagg we're going for 1:1 exact CBOR compatibility with go-fil-markets retrieval types, so things like switching from map encoding to list encoding is not an option. That said, if you want to advocate for pushing cbor-gen to do map ordering PR and use this as an example, happy to push on that.

In the meantime they have to remain the same for wire compatibility.

hannahhoward:
@mvdan there is, as you may have guessed, no schema. These are just the go types as defined by go-fil-markets and encoded with cbor-gen.

However, I can give you the specific examples that will hopefully shed light on what's happening here.

First, often in cbor-gen, pointer types are effectively used to encode optional fields (the obviously conflicts with go-ipld-prime that seperates optional and nullable... not may favorite aspect of the schema design). So for example PieceCID in a retrieval deal proposal is a pointer to a cid, meaning the field is essentially optional. I'd just be happy if we could convert those to nillable fields in a schema and have them work right.

Second, you can see we use cbg.Deferred, a really weird type, for selectors. Basically selectors are ipld.Nodes, and I'd like to have a struct member for a selector that's an ipld.Node and then have a schema with type Any. I guess actually technically selectors have a schema so maybe we could just encode a type schema for them? In case you're wondering, cbg.Deferred is a really weird cbor-gen feature that basically saids "when you get to this field, don't decode it -- just read all the raw byte contents and put them in the field for later decoding" -- it results in some pretty weird code any time you generate retrieval proposals

gammazero (approved):

PR #67, opened 2021-10-20
Link cache is now datastore-backed LRU cache 

### LINKED ISSUE #37: Add cache manager for ingestion in LinkSystem
We currently generate synchronously the entries linked list used for ingestion by triggering the callback in the linksystem if they are not available in the cache. The exchange can't proceed until the full linked list for the entries is generated and stored in the cache.
https://github.com/filecoin-project/indexer-reference-provider/blob/78768e2546a7f81fe7c6237644f5598a83ea5258/core/engine/linksystem.go#L68

The aim of this is to avoid having to keep redundant storages persisting the list of entries advertised. We only generate the linked list for a list of entries when they are requested. However, this process of "generating on-the-fly" is currently quite inefficient.
 https://github.com/filecoin-project/indexer-reference-provider/blob/78768e2546a7f81fe7c6237644f5598a83ea5258/core/engine/linksystem.go#L76

My suggestions to make this faster would be:
- Add a `cacheManager` that keeps track of what has been asked for ingestion, what linked lists need to be generated, and which ones are already in cache and ready to be served. Even more, this would allow us to parallelize the generation of linked lists so at soon as we have the first few nodes of the linked list, we can start serving them using the linked system.
- The `cacheManager` should also include an offline algorithm to garbage collect linkedList (we currently don't have any garbage collection strategy). As a first approach we could  use a naïve approach where we only let a small number of linked list in cache at a time (assuming that after an advertisement publication, all indexers will be looking to ingest the same few set of entries in order to sync).

### LINKED ISSUE #61: Provider daemon shuts down ungracefully due to cancelled context
Engine shutdown call always returns context cancelled error when daemon is stopped via a single `SIGINT`:

```
./provider daemon
2021-10-15T14:17:15.180+0100    INFO    command/reference-provider      provider/daemon.go:87   libp2p host initialized {"host_id": "12D3KooWK9NsyCndVed4QR71wNayaXHcLnRRTq11QPEe8n1w45Jy", "multiaddr": "/ip4/0.0.0.0/tcp/3103"}
2021-10-15T14:17:15.268+0100    INFO    dt-impl impl/impl.go:145        start data-transfer module
2021-10-15T14:17:15.268+0100    INFO    provider/engine engine/engine.go:64     Retrieval address not configured, using /ip4/192.168.68.105/tcp/3103
2021-10-15T14:17:15.268+0100    INFO    command/reference-provider      provider/daemon.go:134  libp2p servers initialized      {"host_id": "12D3KooWK9NsyCndVed4QR71wNayaXHcLnRRTq11QPEe8n1w45Jy"}
2021-10-15T14:17:15.268+0100    INFO    command/reference-provider      provider/daemon.go:154  admin server initialized        {"address": "127.0.0.1:3102"}
Starting admin server on 127.0.0.1:3102 ...2021-10-15T14:17:15.268+0100 INFO    adminserver     http/server.go:58       admin http server listening     {"addr": "127.0.0.1:3102"}
^CReceived interrupt signal, shutting down...
(Hit CTRL-C again to force-shutdown the daemon.)
2021-10-15T14:17:16.584+0100    INFO    pubsub  go-libp2p-pubsub@v0.5.4/pubsub.go:608   pubsub processloop shutting down
2021-10-15T14:17:16.584+0100    INFO    command/reference-provider      provider/daemon.go:170  Shutting down daemon
2021-10-15T14:17:16.584+0100    ERROR   command/reference-provider      provider/daemon.go:184  Error closing provider core     {"err": "context canceled"}
2021-10-15T14:17:16.585+0100    INFO    adminserver     http/server.go:63       admin http server shutdown
2021-10-15T14:17:16.586+0100    INFO    command/reference-provider      provider/daemon.go:199  node stopped
daemon did not stop gracefully
```

I believe the cause is that the context used during shutdown is the parent context from CLI and the [shutdown function does not use the passed context](https://github.com/filecoin-project/indexer-reference-provider/blob/main/engine/engine.go#L159-L161) since underlying go-legs closer does not accept it. 

When shutdown is called the CLI context is already cancelled hence the error - otherwise shutdown would not have been triggered; see: 
* https://github.com/filecoin-project/indexer-reference-provider/blob/main/cmd/provider/daemon.go#L171-L185

gammazero:
This PR uses a datastore-backed LRU cache.  Cache keys (link CIDs) are stored in memory, and the values (multihash chunks) are stored in a persistent datastore.  As new links/chunks are added to the cache, older items are removed when the cache is at capacity.  If a single linked list is more than the cache's current capacity, the cache is resized to hold the entire linked list. 

- Cache size is configurable
- Cache automatically expands to hold largest complete linked list
- Cache persisted across restarts

Fixes #37
Fixes #61

### DIFF
--- a/cmd/provider/daemon.go
+++ b/cmd/provider/daemon.go
@@ -122,7 +122,7 @@ func daemonCommand(cctx *cli.Context) error {
 	}
 
 	// Starting provider core
-	eng, err := engine.New(ctx, privKey, dt, h, ds, cfg.Ingest.PubSubTopic, cfg.ProviderServer.RetrievalMultiaddrs)
+	eng, err := engine.New(ctx, cfg.Ingest, privKey, dt, h, ds, cfg.ProviderServer.RetrievalMultiaddrs)
 	if err != nil {
 		return err
 	}
@@ -176,31 +176,31 @@ func daemonCommand(cctx *cli.Context) error {
 
 	log.Infow("Shutting down daemon")
 
-	ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
 	defer cancel()
 	go func() {
 		// Wait for context to be canceled. If timeout, then exit with error.
-		<-ctx.Done()
-		if ctx.Err() == context.DeadlineExceeded {
-			fmt.Fprintln(cctx.App.ErrWriter, "Timed out on shutdown, terminating...")
+		<-shutdownCtx.Done()
+		if shutdownCtx.Err() == context.DeadlineExceeded {
+			fmt.Println("Timed out on shutdown, terminating...")
 			os.Exit(-1)
 		}
 	}()
 
-	if err := eng.Shutdown(ctx); err != nil {
-		log.Errorw("Error closing provider core", "err", err)
+	if err = eng.Shutdown(shutdownCtx); err != nil {
+		log.Errorf("Error closing provider core: %s", err)
 		finalErr = ErrDaemonStop
 	}
 
 	if err := os.RemoveAll(tmpDir); err != nil {
-		log.Errorw("Error cleaning up temporary files", "err", err)
+		log.Errorf("Error cleaning up temporary files: %s", err)
 		finalErr = ErrCleanupFiles
 	}
 	// cancel libp2p server
 	cancelp2p()
 
-	if err := adminSvr.Shutdown(ctx); err != nil {
-		log.Errorw("Error shutting down admin server", "err", err)
+	if err = adminSvr.Shutdown(shutdownCtx); err != nil {
+		log.Errorw("Error shutting down admin server: %s", err)
 		finalErr = ErrDaemonStop
 	}
 	log.Infow("node stopped")
--- a/engine/dscache.go
+++ b/engine/dscache.go
@@ -1,0 +1,193 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/golang-lru"
+	ds "github.com/ipfs/go-datastore"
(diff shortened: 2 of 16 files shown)


cmd/provider/daemon.go
```diff
@@ -169,31 +169,31 @@ func daemonCommand(cctx *cli.Context) error {
 
 	log.Infow("Shutting down daemon")
 
-	ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
```
willscott, line 172: maybe consider `cctx.Context`?

engine/dscache.go
```diff
@@ -0,0 +1,215 @@
...
+const linksCachePath = "/cache/links"
+
+// dsCache is a LRU-cache that stores keys in memory and values in a datastore.
+type dsCache struct {
+	dstore ds.Datastore
+	lru    *lru.Cache
+	lock   sync.RWMutex
+}
+
+// newDsCache creates a new dsCache instance.  The context is only used cancel
```
willscott, line 23: i feel like we're written very similar caching wrappers around datastores before. i wonder if we can make a canonical one that gets reused.

cmd/provider/daemon.go
gammazero: I specifically did not use `cctx.Context` because that is canceled when the daemon receives CTRL-C, and is used to _trigger_ shutdown.  The `shutdownCtx` is specifically there to give a deadline to complete a graceful shutdown, and if it times out then a hard shutdown follows.

engine/dscache.go
gammazero: I know we have one around the blockstore that uses ARC cache or Two-Queue Cache.  I think cache eviction means something different to it though.  I will take a look to see if I can find others.

PR #72, opened 2021-10-30
Add peer bootstrapping functionality

gammazero:
The indexer can now be configured with bootstrapping peers that the reference provider daemon will attempt to connect to.  This works the same as IPFS bootstrapping.  The minimum number of peers to connect to can also be configured, and if set to 0 will disable bootstrapping.

By configuring bootstrap peer(s), it will be unnecessary to use the `connect` command to connect to an indexer.


willscott (approved):

masih:
Nice 👍 any reason to keep `connect` now that there is bootstrapping?

PR #87, opened 2021-11-08
Simplify admin server error response

### LINKED ISSUE #80: Simplify HTTP server responses
The admin server delivers error responses using an `errRes` data structure that gets serialized by the `respond()` function.  This was originally to have a common way of delivering errors over both libp2p and http, but is not needed here.

Here is an example of the current error response:
```go
errRes := newErrorResponse("failed to supply CAR. %v", err)                                                                                                                             
respond(w, http.StatusInternalServerError, errRes)                                                                                                                                       
```
I would prefer to see:
```go
http.Error(w, fmt.Sprintf("failed to supply CAR: %s", err), http.StatusInternalServerError)
```
Even that can be refined, so that if the file is not found it returns `http.StatusBadRequest`

masih:
Use a simple `http.Error` when responding with error in admin server.
Reflect changes in the CLI so that the response body is displayed as
expected.

Use consistent way to construct http error in admin server, and reuse
the log message in http response.

Add utility to interact with admin HTTP server in CLI commands to reduce
duplicate code. Refactor CLI commands to use the utilities for both
parsing of error response from admin server and sending of HTTP
requests.

Fixes #80


gammazero (approved):
Nice.  Condensed a bunch of common code.

PR #93, opened 2021-11-15
split cmd/ into a separate module

mvdan:
(see commit message)


mvdan:
FYI @dirkmc

mvdan:
CI should still work, thanks to the multiple-go-modules action - it should still be testing all modules.

It looks like I messed up the placement of some testdata files, though.

cmd/provider/daemon.go
```diff
@@ -11,12 +11,12 @@ import (
...
-	"github.com/filecoin-project/indexer-reference-provider/config"
-	"github.com/filecoin-project/indexer-reference-provider/engine"
-	"github.com/filecoin-project/indexer-reference-provider/internal/cardatatransfer"
-	"github.com/filecoin-project/indexer-reference-provider/internal/suppliers"
-	adminserver "github.com/filecoin-project/indexer-reference-provider/server/admin/http"
-	p2pserver "github.com/filecoin-project/indexer-reference-provider/server/provider/libp2p"
+	"github.com/filecoin-project/indexer-reference-provider/lib/config"
+	"github.com/filecoin-project/indexer-reference-provider/lib/engine"
+	"github.com/filecoin-project/indexer-reference-provider/lib/cardatatransfer"
+	"github.com/filecoin-project/indexer-reference-provider/lib/suppliers"
```
gammazero, line 17: Maybe rename this `supplier` to follow the no plurals package naming convention

gammazero (commented):
If this project is primarily a lib, it seems like a `/lib` directory is superfluous.  I suggest considering that every exported package live in its own directory under root.  Things like `cmd/` will be recognized to be special.

Either way, this is an important consideration since ti establishes repo organizational patterns that will be duplicated in other repos.

mvdan:
> If this project is primarily a lib, it seems like a `/lib` directory is superfluous.

From a conceptual level it's certainly unnecessary, but from a practical level I want to avoid overlapping Go modules. That is, if we have `/go.mod` and `/cmd/go.mod`, you end up with "nested Go modules", which make using them more confusing. For example, if someone does a `go get` of `[...]/index-provider/cmd/provider`, that's an ambiguous import path; the Go tool must check if either the `index-provider` or `index-provider/cmd` modules provide such a package. It could be none of them (causing an error), just one of them, or both of them (also causing an error).

Another example where we can cause user confusion is that the module zip for the root module won't include the nested module directory. This is fairly obvious if the two modules aren't overlapping, but if they are, you could be forgiven for being bitten by this detail.

We could certainly avoid the `/lib` prefix and use nested Go modules, it's not impossible. It would add sharp edges, though. One useful data point is that in go-car we did end up with `/go.mod` and `/cmd/go.mod`, but note that the actual latest library is under `/v2/go.mod`, so that still doesn't overlap with `/cmd`.

I'm not attached to the name `/lib` either. If we want to keep the non-overlapping modules but use a different name, that's fine too.

> Either way, this is an important consideration since ti establishes repo organizational patterns that will be duplicated in other repos.

I wouldn't say so, in general. It should be relatively rare for repositories to need multiple modules.

mvdan:
briefly discussed in colo - since we've used overlapping/nested modules in other repos like go-car and go-ipld-prime, and they haven't caused much fuss, we're going to be practical and drop the `/lib` prefix.

cmd/provider/daemon.go
mvdan: Done!

mvdan:
Undone the move to `lib/`. Should be ready to merge now.

@masih I'm leaving the makefile and dockerfile to you for a follow-up PR, since you volunteered :)

mvdan:
Huh, I'm really not having a good day today with Go master. Found another bug while working on this PR. https://github.com/golang/go/issues/49598

mvdan:
Hmm, how many times can one get `go mod tidy` and gofmt wrong? Apologies for the noise. Fingers crossed it's green now.

go.mod
```diff
@@ -11,27 +11,30 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.7.4
+	github.com/ipfs/go-bitswap v0.4.0 // indirect
```
gammazero, line 14: Seems like it should be OK to remove the indirect packages.

gammazero (approved):
I would prefer to also git rid of the `cmd/priovider/` directory and put everything there into `cmd/`.  That can be done in a separate PR if preferable.

go.mod
mvdan: not sure I follow; this is a tidied module. we could possibly remove this line to use an older version of bitswap, if that's what you mean, but presumably that would be a change compared to what we had before.

masih (approved):

mvdan:
> I would prefer to also git rid of the `cmd/priovider/` directory and put everything there into `cmd/`. That can be done in a separate PR if preferable.

I'd keep the sub-folder, because it's possible we will have other commands later on. It's also nice when one can do `go install ./cmd/provider` and the binary name ends up being `provider`; otherwise one ends up with a binary named `cmd` by default.

go.mod
mvdan: Another reason to keep this as-is is that this keeps the ipfs-related dependencies on the same versions in both the cmd and root modules, which will simplify testing and make our software more consistent. Otherwise we'd be testing and using two sets of dependency versions, and I don't think we want to go there.

Once 1.18 is out, we can add a go.work file to help keep those shared dep versions in sync. For now, I'll leave this as-is.

PR #95, opened 2021-11-16
Downgrade dependencies to reduce churn in lotus integration

masih:
Downgrade following dependencies to reduce churn on lotus side:
 - github.com/libp2p/go-libp2p-quic-transport to `v0.11.2`
 - github.com/ipfs/go-bitswap to `v0.3.4`

Use `replace` directive in `cmd/go.mod` so that the latest changes from
root module are pulled.

Relates to:
 - https://github.com/filecoin-project/lotus/pull/7622

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.7.4
-	github.com/ipfs/go-bitswap v0.4.0 // indirect
 	github.com/ipfs/go-cid v0.1.0
 	github.com/ipfs/go-datastore v0.4.6
 	github.com/ipfs/go-ds-leveldb v0.4.2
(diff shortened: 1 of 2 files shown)


cmd/go.mod
```diff
@@ -22,3 +22,5 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/urfave/cli/v2 v2.3.0
 )
+
+replace github.com/filecoin-project/index-provider => ../
```
willscott, line 26: i don't think you want this, right?

mvdan: I think we do - makes maintenance on our side easier. The only reason we would want to avoid a replace directive is if this module was a library, in which case we'd want to make sure that our users don't need to copy-paste the same replace directives.

mvdan (approved):

cmd/go.mod
masih: The issue I see around the corner is having to roll out big changes in cmd module when root module changes are piled up. Removing from this PR since it's a separate issue and beyond the scope of this PR so that we can roll out the changes needed by ignite

masih: ah saw @mvdan's message after my comment adding it back in

gammazero: I do not think we want this.  Instead, how about:
https://github.com/filecoin-project/index-provider/pull/97

masih: OK I am going to remove replace directive from this PR so that we can push ahead with dependency downgrade and avoid blocking ignite. In another PR we can then continue the discussion re `replace` directive.

masih: The discussion to be continued elsewhere probably in #97. Resolving conversation.

PR #99, opened 2021-11-17
Do not close the wrapped datastore used for caching

masih:
The index provider engine as well as its internal components, like
`dsCache` take an existing datastore instance and reuse it internally.
`dsChace` implements the `datastore.Datastore` interface by wapping the
existing instance. Since the instance is instantiated externally, it
should not be closed off by the provider `Engine` nor `dsChache`.

Instead, delegate the management of the datastore lifecycle to where it
is instantiated, i.e. the `daemon` command.

### DIFF
--- a/engine/dscache.go
+++ b/engine/dscache.go
@@ -20,6 +20,8 @@ type dsCache struct {
 	lock   sync.RWMutex
 }
 
+var _ ds.Datastore = (*dsCache)(nil)
+
 // newDsCache creates a new dsCache instance.  The context is only used cancel
 // a call to this function while it is accessing the data store.
 func newDsCache(ctx context.Context, dstore ds.Datastore, capacity int) (*dsCache, error) {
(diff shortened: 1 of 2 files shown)


dirkmc (approved):

aarshkshah1992 (approved):

engine/dscache.go
```diff
@@ -110,19 +112,11 @@ func (c *dsCache) Sync(key ds.Key) error {
...
-
-	if c.dstore == nil {
-		return nil
-	}
-
-	if err := c.dstore.Close(); err != nil {
-		return err
-	}
-	c.dstore = nil
 	return nil
```
gammazero, line 120: I think that it is at least necessary to do a `c.dstore.Sync()` here as it may be expected that things are persisted to disk after close.

aarshkshah1992: I agree.

masih: Good call thank you

gammazero (approved):
Do you think it would be better to open a different datastore?

masih:
> Do you think it would be better to open a different datastore?

That would work too; I am not sure how much it matters. A potential reason to separate it long-term may be to be able to mount different type of storage for caching for performance reasons. But then I'd love to see some data showing we need faster cache before putting in the effort.

PR #111, opened 2021-12-14
Remove libp2p server exposed by provider and rewrite e2e tests

### LINKED ISSUE #110: Remove now redundant libp2p server from provider
Remove the[ libp2p server ](https://github.com/filecoin-project/index-provider/blob/main/server/provider/libp2p/handler.go#L36-L71)exposed by the provider, because the server predates "go-legs" and what it provides is now satisfiable via alternative already exposed endpoints:
* `GET_LATEST` is now exposed via go-legs' head publisher over go stream
* `GET_AD` is available via graphsyc.

The server is used in [end to end tests](https://github.com/filecoin-project/index-provider/blob/main/e2e_retrieve_test.go) which needs to be refactored to use graphsyc instead. 

Once removed, the client library for it can also go away [here](https://github.com/filecoin-project/storetheindex/tree/main/providerclient/libp2p).

Note that storetheindex uses the libp2p[ to get the latest](https://github.com/filecoin-project/storetheindex/blob/main/internal/ingest/ingest.go#L208) advertisement instead of go-legs head publisher which also needs to be updated, captured in https://github.com/filecoin-project/storetheindex/issues/137

masih:
Remove the libp2p server in provider in favour of go-legs since it
already provides the functionality of that server.

The sever was used in e2e testing of car data transfer. Rewrite the
tests to use go-legs graphsyc endpoint for fetching advertisements. Tidy
up tests as needed while at it.

Fixes #110


willscott (approved):
yay code removal

PR #112, opened 2021-12-15
Refactor config for easier instantiation with default values

### LINKED ISSUE #108: Expose the ability to construct sub config with default values
The only way to make a config construct with defaults just now is to to initialise one with identity, take the sub config needed and throw the rest away. Default values are not exported.

Expose the ability to construct an instance of a config/sub-config with default values set.

masih:
Add `New*` constructor for each of the sub config types with default
values.

Fixed #108

### DIFF
--- a/config/adminserver.go
+++ b/config/adminserver.go
@@ -20,6 +20,15 @@ type AdminServer struct {
 	WriteTimeout    Duration
 }
 
+// NewAdminServer instantiates a new AdminServer config with default values.
+func NewAdminServer() AdminServer {
+	return AdminServer{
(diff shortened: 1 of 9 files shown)


gammazero (approved):
LGTM, with one minor comment.  Do you prefer keeping the const values or removing them like was done with storetheindex config?

masih:
> Do you prefer keeping the const values or removing them like was done with storetheindex config

In this case I prefer keeping them _because none are exported_. If you think there is a danger of them getting exported at some point I'm more than happy to remove :)

Issue #121, opened 2022-01-03
removal generates invalid advertisement


willscott:
when removing by metadata context ID:

```
2022-01-03T08:52:29.283Z        INFO    provider/engine engine/engine.go:331    Generating removal list for advertisement
2022-01-03T08:52:29.287Z        ERROR   provider/engine engine/engine.go:380    Error generating new advertisement: storetheindex: invalid metadata: encountered protocol ID 0 on encode
2022-01-03T08:52:29.287Z        ERROR   adminserver     http/removecar_handler.go:47    failed to remove CAR: storetheindex: invalid metadata: encountered protocol ID 0 on encode      {"err": "storetheindex: invalid metadata: encountered protocol ID 0 on encode", "key": "NQe/ch+eG2aP3DE1xSMcbGF+mJGLH6WmiA8z4xB5QXg="}
```

gammazero:
Fixed by #124

PR #124, opened 2022-01-06
Generate valid removal advertisements

gammazero:
Advertisements require valid metadata, even though metadata is not used for index removal.  This change creates a removal advertisement with valid, but empty, metadata.

Additional changes:
- Fix bad contextID creation in remove command
- Improve output of find, import, and remove commands
- Improve log messages: readable contextID, structured logging

Fixes issue #121

### DIFF
--- a/cmd/provider/remove.go
+++ b/cmd/provider/remove.go
@@ -44,25 +44,25 @@ specified, they key is simply calculated as the SHA_256 hash of the given path.`
 )
 
 func beforeRemoveCar(cctx *cli.Context) error {
-	keyFlagSet := cctx.IsSet(keyFlag.Name)
-	carPathFlagSet := cctx.IsSet(carPathFlag.Name)
-
-	if keyFlagSet && carPathFlagSet {
-		return fmt.Errorf("only one of %s or %s must be set", keyFlag.Name, carPathFlag.Name)
-	}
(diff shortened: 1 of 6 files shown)


cmd/provider/remove.go
```diff
@@ -44,25 +44,25 @@ specified, they key is simply calculated as the SHA_256 hash of the given path.`
...
-		if err != nil {
-			return errors.New("key is not a valid base64 encoded string")
-		}
-		removeCarKey = decoded
-	} else {
-		removeCarKey = sha256.New().Sum([]byte(carPathFlagValue))
+	if cctx.IsSet(optionalCarPathFlag.Name) {
+		return fmt.Errorf("only one of %s or %s must be set", keyFlag.Name, optionalCarPathFlag.Name)
+	}
+	decoded, err := base64.StdEncoding.DecodeString(keyFlagValue)
```
willscott, line 61: can we have usage for this flag indicate it should be base64 encoded when passed in?

gammazero: The usage for `keyFlag` indicates base64:
```go
Usage:       "Base64 encoded lookup key to associate with imported CAR."
```
https://github.com/filecoin-project/index-provider/blob/main/cmd/provider/flags.go#L122

The `optionalCarPathFlag` is a path so should not be base64 encoded.

PR #125, opened 2022-01-07
Use new go-datastore with contexts

gammazero:
This requires updates to many other dependencies.

Other changes:
- Removed mostly unused utils/utils.go
- Replaced engine/dscache.go with enging/lrustore/lrustore.go


Makefile
```diff
@@ -1,16 +1,17 @@
...
+build: 
+	cd $(BIN_SUBDIR) && go build
 
 docker: Dockerfile clean
 	docker build . --force-rm -f Dockerfile -t indexer-reference-provider:$(shell git rev-parse --short HEAD)
 
-$(BIN): vet test
-	cd cmd/provider && go build -o ../../$(@)
+install:
+	cd $(BIN_SUBDIR) && go install
```
willscott, line 14: I generally prefer build of a binary over `go install` which can often put the binary into an unexpected location.

willscott (approved):
this looks largely mechanical apart from the change to the makefile

Makefile
gammazero: There is still a `build` target.  The biggest difference is that it does not move the executable command into the repo root directory, but keeps it in the cmd/provider directory.  This was done to provide what is generally more expected behavior.  The `install` target is simply convenience.

PR #126, opened 2022-01-10
Re-enable flaky tests and remove `time.Sleep` s using event tracer

### LINKED ISSUE #12: Fix flaky tests
The goal of this issue is to track the list of unit tests that are flaky and need some extra care to make it work:
_If you see any other flaky test that needs to be investigated please add it to the list_
- [ ] `TestNotifyPublish` from `engine_test.go` (fails in MacOS go 1.17)
     - This one is probably due to the use of `time.Sleep` instead of an async approach for tests. It fails in `go test -race` with runs slower than vanilla `go test`

masih:
Now that go-legs is heavily refactored, re-enable the flaky tests and
remove explicit `time.Sleep` calls in the tests. This is so that we can
check if tests remain flaky and if so how to fix them so that they are
not.

Implement a bespoke pubsub event tracer that allows tests to observe and
wait for pubsub events before proceeding with other assertions.

Refactor assertions to reduce duplicate code across engine tests.

 Fixes #12

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -76,6 +76,10 @@ type Engine struct {
 	// cb is the callback used in the linkSystem
 	cb   provider.Callback
 	cblk sync.Mutex
+
+	// dtsyncOptions specifies the options used to instantiate legs dtsync.publisher on Engine.Start.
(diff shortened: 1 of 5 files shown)


masih:
Fixed in #193

PR #129, opened 2022-01-12
Update godoc across packages

masih:
Add missing documentation and alter as needed.

### DIFF
--- a/cmd/provider/doc.go
+++ b/cmd/provider/doc.go
@@ -1,0 +1,64 @@
+/*
+
+Provider CLI can be used to start an index-provider daemon that runs a reference implementation of
+the provider interface, with ability to publish advertisements by importing/removing CAR files
+and server content retrieval via DataTransfer GraphSync protocol.
+
+Usage:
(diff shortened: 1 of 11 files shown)


cmd/provider/doc.go
```diff
@@ -0,0 +1,64 @@
...
+
+Usage:
+	NAME:
+	   provider - Indexer Reference Provider Implementation
+
+	USAGE:
+	   provider [global options] command [command options] [arguments...]
+
+	VERSION:
+	   v0.0.0+unknown
```
willscott, line 15: maybe don't need this - it'll just always be outdated

willscott (approved):

cmd/provider/doc.go
masih: yup; none of the build commands set that var as far as I can tell so it's always unknown I _think_. Will capture a separate ticket.

masih: https://github.com/filecoin-project/index-provider/issues/130

PR #139, opened 2022-01-13
Support http publisher

MarcoPolo:
## Context

go-legs supports syncing over http. But the index provider hasn't been hooked up to start the http server.

## Changes

This PR adds a config and logic to enable the http index provider via go-legs. It also includes a test in `e2e_retrieve_test.go` that syncs over http.

### DIFF
--- a/config/httpserver.go
+++ b/config/httpserver.go
@@ -1,0 +1,34 @@
+package config
+
+import (
+	"github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+type HttpServer struct {
+	Enabled         bool
+	ListenMultiaddr string
+}
+
+// NewHttpServer instantiates a new config with default values.
+func NewHttpServer() HttpServer {
+	return HttpServer{
+		Enabled:         false,
+		ListenMultiaddr: "/ip4/0.0.0.0/tcp/3104/http",
+	}
+}
+
+func (hs *HttpServer) ListenNetAddr() (string, error) {
+	maddr, err := multiaddr.NewMultiaddr(hs.ListenMultiaddr)
+	if err != nil {
+		return "", err
+	}
+	httpMultiaddr, _ := multiaddr.NewMultiaddr("/http")
+	maddr = maddr.Decapsulate(httpMultiaddr)
+
+	netAddr, err := manet.ToNetAddr(maddr)
+	if err != nil {
+		return "", err
+	}
+	return netAddr.String(), nil
+}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -56,8 +56,10 @@ type Engine struct {
 	cache     datastore.Datastore
 	// ds is the datastore used for persistence of different assets (advertisements,
 	// indexed data, etc.)
-	ds        datastore.Batching
-	publisher legs.Publisher
+	ds datastore.Batching
+
+	publisherKind config.PublisherKind
+	publisher     legs.Publisher
 
 	httpPublisherCfg *config.HttpPublisher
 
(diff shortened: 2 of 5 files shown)


engine/engine.go
```diff
@@ -212,7 +231,19 @@ func (e *Engine) Publish(ctx context.Context, adv schema.Advertisement) (cid.Cid
 
 	log.Infow("Publishing advertisement in pubsub channel", "cid", c.String())
 	// Use legPublisher to publish the advertisement.
-	return c, e.publisher.UpdateRoot(ctx, c)
+	err = e.publisher.UpdateRoot(ctx, c)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	if e.httpPublisher != nil {
```
willscott, line 239: can we have the publisher be one or the other, rather than both a graphsync and http publisher?

config/httpserver.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	httpMultiaddr, _ := multiaddr.NewMultiaddr("/http")
+	maddr = maddr.Decapsulate(httpMultiaddr)
+
+	netAddr, err := manet.ToNetAddr(maddr)
+	if err != nil {
+		return "", err
+	}
+	return netAddr.String(), nil
+}
```
gammazero, line 34: This seems like the wrong package for this file.  The config package should only be concerned about settings read from the config file. 

I think that `ListenMultiaddr` [here](https://github.com/filecoin-project/index-provider/blob/main/config/providerserver.go#L5) should used.  If it is an HTTP address then start the http publisher, otherwise start the datatransfer publisher.  The rest should go under `server/`.

gammazero (approved):

engine/engine.go
MarcoPolo: done

config/httpserver.go
```diff
@@ -0,0 +1,34 @@
+package config
+
+import (
+	"github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+type HttpServer struct {
+	Enabled         bool
```
masih, line 9: Please document behaviour of this flag; enabling HTTP publisher seems to disable dtsync publisher?

engine/engine.go
```diff
@@ -161,11 +165,23 @@ func (e *Engine) Start(ctx context.Context) error {
 
 	e.cache = dsCache
 
-	e.publisher, err = dtsync.NewPublisherFromExisting(e.dataTransfer, e.host, e.pubSubTopic, e.lsys)
+	if e.httpPublisherCfg.Enabled {
```
masih, line 168: So means Provider would start either a HTTP publisher or `dtsync.Publusher`? 

I think we need to document this.

engine/engine.go
```diff
@@ -258,11 +278,13 @@ func (e *Engine) Shutdown() error {
 	err := e.publisher.Close()
 	if err != nil {
 		err = fmt.Errorf("error closing leg publisher: %w", err)
+		return err
```
masih, line 281: This means `e.cache` will not close if we fail to close the publisher. 

I am curious why this change?

MarcoPolo: ah yes, I mean to undo this. Thanks for the catch!

engine/engine.go
MarcoPolo: Changed in 0a8a8bb1697d77109f30f20627fc034e77da98f4 with an enum field in the ingest cfg.

config/httpserver.go
```diff
@@ -0,0 +1,34 @@
+package config
+
+import (
+	"github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+type HttpServer struct {
```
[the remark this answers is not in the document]
MarcoPolo: Done 336fa9deca0b78cfcdab22f7b75bc7f6b0d16772

config/httpserver.go
MarcoPolo: Changed via enum: 0a8a8bb1697d77109f30f20627fc034e77da98f4

MarcoPolo:
Great feedback @masih! Updated with your suggestions

gammazero (approved):
LGTM

PR #141, opened 2022-01-14
Implement a CLI command to verify multihash ingestion

### LINKED ISSUE #137: Implement a utility for verifying that expected multihashes are ingested by `storetheindex`
Write a utility that verifies multihashes are known by indexer node , more specifically `storetheindex` implementation of it.

The utility is to be used for verification in miners in its first iteration. It will accept as input: 
* miner peer ID
* CAR index files 
* storetheindex endpoint

and for each multihash found in the index checks that `storetheindex` has it associated to a provider with the same peer ID.

The tool should also be able to get the list of advertisements from the provider's graphsync API as a source of multihashes to verify.

masih:
Implement a CLI tool that given a source of multihashes it verifies that
every multihash is indexed by an indexer node and is associated to the
expected peer ID as its provider. The sources of multihash may be CAR
files, CARv2 iterable multihash index files or Lotus markets dagstore
repo (migrated to the latest CAR index format).

This utility is written to aid miners verify their deals are indexed
by an indexer node.

Relates to #137


.gitignore
```diff
@@ -1,4 +1,4 @@
 *~
 index-provider
 cmd/provider/provider
-provider
+./provider
```
masih, line 4: Otherwise ignores any director with name `provider`

willscott:
cool. it'll be probably useful to have some way to sample, or say like 'do every 100th cid'  - there may be a lot of CIDs, so getting a sample, eventually even just 1 or 2 per deal to understand which deals seem indexed will be useful.

gammazero (approved):

willscott:
other question is if rather than having this be dagstore / lotus specific, can't we just look at the index provider advertisement chain directly to find CIDs? there's nothing in this utility that _needs_ to be lotus specific, i think

masih:
Outstanding things for me:
* The max HTTP message size on storetheindex; the call to find batch always uses all the multihashes found in an index file, and I have no idea how big the number can be. Any thoughts? Would some rough max 16K multihashes per findBatch call to indexer make sense? 
* If checking fails half way in the middle because say indexer is redeployed or connection error or some corrupt file, should the tool support resumption?
* Do we need detailed logging of shard keys or deals Multihashes belog to for the ones that fail verification?

> cool. it'll be probably useful to have some way to sample, or say like 'do every 100th cid' - there may be a lot of CIDs, so getting a sample, eventually even just 1 or 2 per deal to understand which deals seem indexed will be useful.

I see; as in partial verification rather than full verification? Sure, sounds good. 

> rather than having this be dagstore / lotus specific, can't we just look at the index provider advertisement chain directly to find CIDs? there's nothing in this utility that needs to be lotus specific, i think

Great point. For this feature the tool will then depend on a running provider and would need a way to walk the chain of advertisements. This means getting the chain is either exposed via admin API or the tool makes graphsync calls to the provider to get it. Unless the provider datastore is accessed directly?
Any thoughts?

gammazero:
> This means getting the chain is either exposed via admin API or the tool makes graphsync calls to the provider to get it. Unless the provider datastore is accessed directly?
Any thoughts?

I think the tool should make graphsync requests so that is can be used against any number of providers, regardless of their datastore implementation.

PR #142, opened 2022-01-15
Tell indexer to remove content by context ID

gammazero:
This change explicitly tells the indexer to remove content by context.  It does this by creating a removal ad that has no links and instead uses the special NoEntries link.

Depends on https://github.com/filecoin-project/storetheindex/pull/155

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -428,16 +428,19 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, metad
 			return cid.Undef, err
 		}
 
-		// Provide the CID link to the content entries to give the indexer the
-		// option to remove all individual entries.  If a contextID is given
-		// then the indexer will delete all content for that contextID and not
-		// need to retrieve the entries.
-		cidsLnk = cidlink.Link{Cid: c}
+		// Create an advertisement to delete content by contextID by specifying
+		// that advertisement has no entries.
+		cidsLnk = schema.NoEntries
+
+		// To delete specific indexes values, provide the CID link to the
+		// content entries to delete.  The indexer will fetch these entries and
+		// delete indexes for the content in each entry chunk.
+		//   cidsLnk = cidlink.Link{Cid: c}
 
 		// The advertisement still requires a valid metadata even though it is
 		// not used for removal.  Create a valid empty metadata.
 		metadata = stiapi.Metadata{
-			ProtocolID: cardatatransfer.ContextIDCodec, //providerProtocolID,
+			ProtocolID: cardatatransfer.ContextIDCodec,
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


engine/engine.go
```diff
@@ -428,16 +428,19 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, metad
...
-		// need to retrieve the entries.
-		cidsLnk = cidlink.Link{Cid: c}
+		// Create an advertisement to delete content by contextID by specifying
+		// that advertisement has no entries.
+		cidsLnk = schema.NoEntries
+
+		// To delete specific indexes values, provide the CID link to the
+		// content entries to delete.  The indexer will fetch these entries and
+		// delete indexes for the content in each entry chunk.
+		//   cidsLnk = cidlink.Link{Cid: c}
```
masih, line 438: why keep this line as commented out?

masih, line 438: Why keep this line commented out?

engine/engine.go
```diff
@@ -428,16 +428,19 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, metad
...
-		// Provide the CID link to the content entries to give the indexer the
-		// option to remove all individual entries.  If a contextID is given
-		// then the indexer will delete all content for that contextID and not
-		// need to retrieve the entries.
-		cidsLnk = cidlink.Link{Cid: c}
+		// Create an advertisement to delete content by contextID by specifying
+		// that advertisement has no entries.
+		cidsLnk = schema.NoEntries
+
+		// To delete specific indexes values, provide the CID link to the
```
masih, line 435:
```suggestion
		// To delete specific index values, provide the CID link to the
```

engine/engine.go
```diff
@@ -428,16 +428,19 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, metad
...
 		}
 
-		// Provide the CID link to the content entries to give the indexer the
-		// option to remove all individual entries.  If a contextID is given
-		// then the indexer will delete all content for that contextID and not
-		// need to retrieve the entries.
-		cidsLnk = cidlink.Link{Cid: c}
+		// Create an advertisement to delete content by contextID by specifying
+		// that advertisement has no entries.
+		cidsLnk = schema.NoEntries
```
masih, line 433: Could you help me understand the rationale behind this change please?
Does this change not assume that ContextID is unique per advertisement? Otherwise, this would wipe all the indices associated to the ContextID; no?

Am I right to assume ContextID [identifies metadata](https://github.com/filecoin-project/go-indexer-core/blob/main/value.go#L15) and not advertisement?

engine/engine.go
gammazero:
```suggestion
```

engine/engine.go
gammazero: When indexed content associated with a context ID, to delete specific indexes then advertise entries that specify the indexes to delete.  When deleting _all_ content associated with a context ID, explicitly do not include entries.

This change does not make any assumption about whether a context ID is unique to an advertisement.

Yes, this change will wipe out all the indices associated with a context.  The comment specifies how to do otherwise.

masih (approved):

PR #147, opened 2022-01-21
Implement command to list provider advertisement

masih:
Implement a CLI command that gets the content of an advertisement from
a provider endpoint either via HTTP or over graphsync depending on the
given endpoint with option to print the entire list of multihashes.

### DIFF
--- a/cmd/provider/get.go
+++ b/cmd/provider/get.go
@@ -1,0 +1,129 @@
+package main
+
+import (
+	"fmt"
+	"github.com/filecoin-project/index-provider/cmd/provider/internal"
+	"github.com/ipfs/go-cid"
+	"github.com/libp2p/go-libp2p-core/peer"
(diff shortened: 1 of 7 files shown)


cmd/provider/get.go
```diff
@@ -0,0 +1,129 @@
...
+				Destination: &printEntries,
+			},
+		},
+	}
+)
+
+func beforeGetAdvertisements(cctx *cli.Context) error {
+	var err error
+	if cctx.NArg() > 1 {
+		return cli.Exit("At most one argument [ad-cid] must be specified. If non is specified, the current head advertisement is fetched.", 1)
```
willscott, line 56:
```suggestion
		return cli.Exit("At most one argument [ad-cid] must be specified. If none is specified, the current head advertisement is fetched.", 1)
```

willscott (approved):

PR #148, opened 2022-01-21
Add CLI command to verify ingestion by getting ad directly from provider

masih:
Add `provider verify-ingest` option to get the advertisement by directly
connecting to a provider's grphsync endpoint and fetching either the
latest advertisement entries or entries belonging to a specific
advertisement CID.

### DIFF
--- a/cmd/provider/doc.go
+++ b/cmd/provider/doc.go
@@ -24,6 +24,7 @@ Usage:
 	   register           Register provider information with an indexer that trusts the provider
 	   remove, rm         Removes previously advertised multihashes by the provider.
 	   verify-ingest, vi  Verifies ingestion of multihashes to an indexer node from a Lotus miner, CAR file or a CARv2 Index
+	   list               Lists advertisements
 	   help, h            Shows a list of commands or help for one command
 
 	GLOBAL OPTIONS:
--- a/cmd/provider/internal/entries_iter.go
+++ b/cmd/provider/internal/entries_iter.go
@@ -47,7 +47,7 @@ func (d *EntriesIterator) Next() (multihash.Multihash, error) {
 	return d.chunkIter.Next()
 }
 
-func (d *EntriesIterator) Drain() ([]multihash.Multihash, int, error) {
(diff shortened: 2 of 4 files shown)


cmd/provider/doc.go
```diff
@@ -24,6 +24,7 @@ Usage:
 	   register           Register provider information with an indexer that trusts the provider
 	   remove, rm         Removes previously advertised multihashes by the provider.
 	   verify-ingest, vi  Verifies ingestion of multihashes to an indexer node from a Lotus miner, CAR file or a CARv2 Index
+	   get-ad             Gets advertisement
```
willscott, line 27: maybe `list`?

cmd/provider/get.go
```diff
@@ -62,41 +64,30 @@ func beforeGetAdvertisements(cctx *cli.Context) error {
...
-	} else {
-		if topic == "" {
-			return cli.Exit("topic must be configured when graphsync endpoint is specified.", 1)
-		}
-		provClient, err = internal.NewGraphSyncProviderClient(addrInfo, topic)
-		if err != nil {
-			return err
-		}
+	if topic == "" {
+		return nil, errors.New("topic must be configured when graphsync endpoint is specified")
```
willscott, line 88: why is this? i thought topic was just for the gossiping - is it also used as the name of the `head` service?

masih: Yes for the head.

willscott: there is i believe a way to enumerate supported protocols with multistream. i wonder if there's a sane default heuristic that we can use for guessing which one it probably is?

masih: My suggestion would be to update that in go-legs and reflect here. Cause right now creating a new syncer requires topic: https://github.com/filecoin-project/go-legs/blob/main/dtsync/sync.go#L103

Which I think is used only for head query.

willscott (approved):

cmd/provider/get.go
willscott: fair enough - though i guess the use in diagnostics is different from the main protocol where it should be known in advance

PR #163, opened 2022-01-31
Update to storetheindex v0.3.0

gammazero:
- Updates to the latest storetheindex
 - Fixes default codec value in index command


cmd/provider/index.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+const defaultMetadataCodec multicodec.Code = 0x300001
```
willscott, line 16: comment on what this is

willscott (approved):
does this also bump to the smaller signatures?

gammazero:
Changing over to smaller signatures was done in a previous PR which uses a non-tagged commit of storetheindex.  This gets the storetheindex dependency back to a tagged commit.

Issue #118, opened 2022-01-02
Import ux slow / needs improvment


willscott:
```
$ ./index-provider import car -l http://localhost:3102 -i bafybeihh.carv2
Post "http://localhost:3102/admin/import/car": EOF
```

* this is a carv2, but the length of time for import seems to be the same as for a carv1
* the 'EOF' seems not like a useful response

masih:
Do you know:
* What was the index codec of `bafybeihh.carv2`?
* Did it have  [fully indexed](https://github.com/ipld/go-car/blob/master/v2/car.go#L72) characteristic set?

willscott:
It was made using `car index`, with the [default codec](https://github.com/ipld/go-car/blob/master/cmd/car/car.go#L112)

do we have any tools that will cause the fully indexed characteristic to be set? should we?

willscott:
the 'isfullyindexed' characteristic makes no sense to me. it seems to not be a property of a car, but the context in which the car is being used, right? the same car is both fully index and not fully indexed depending not in it's contents, but on whether someone thought identity CIDs might be there?

masih:
Fully indexed characteristic simply means does the index represent a catalog of all CIDs in the CAR or not, so I'd say it does depend on the content of the CAR. For example, a CAR can be indexed with an option to ignore IDENTITY CIDs. The resulting index would not then be a catalog of all CIDs in the CAR and will not have "fully indexed" characteristic. 

For the purposes of index-provider what we do want is a catalog of all CIDs in a CAR, i.e. need to also advertise IDENTITY CIDs. If a given CAR index is not a full catalog then we'd need to re-generate the index.

A small note on where all of this comes from: maintaining backward compatibility while supporting CAR indices that do include IDENTITY CIDs. For context see https://github.com/ipld/go-car/pull/239 and https://github.com/ipld/ipld/pull/138

willscott:
if i have a car i'm generating an index for, i don't know if it was meant to have identity cids, but didn't or if it just didn't have identity cids. in what contexts are we comfortable setting the bit?

masih:
I see what you are saying. It all depends on the use case. In FileCoin use case IDENTITY CIDs should be included because I think otherwise the commp calculation will not match among other reasons (e.g. IDENTITY CIDs with link in data block). Remember, these options were added after the effect to avoid breaking changes where existing usage didn't match the spec at the time.

What go-car is trying to do is to facilitate usage, may it be including IDENTIY or not including it. 

[...]

PR #164, opened 2022-01-31
Relax the need for full index in car supplier

masih:
The CAR supplier regenerated the index if it did not contain identity
CIDs, i.e. the fully-indexed characteristic was missing in CARv2 header.

See: https://github.com/filecoin-project/index-provider/issues/118

Fixes: #118


willscott (approved):
🙇

Issue #73, opened 2021-10-31
if car supplier gets a carv1 it should save an index for it (as the dagstore would do)


willscott:
currently a restart would need to re-scan the full carv1 if a retrieval or callback for it's index is ever encountered.

masih:
Where would the index be stored? Are you thinking car supplier would modify the CARv1?

willscott:
the reference provider maintains a data store with it's chain of advertisements. presumably in the same place?
probably safer to keep it as a detached index rather than modifying the source file directly.

masih:
On a separate but related subject, I wonder if it makes sense to also store the deterministically ordered index, regardless of CAR version.

gammazero:
The linksystem cache is already doing this, although not indefinitely. This is persisted across restarts now that the cache is backed by a data store, since PR #67.

The linked chunks for each advertisement are cached when the advertisement is published.  When the indexer asks for the advertised index data, the provider serves it from cache.  Being a cache, after some number of other advertisements are generated, older data may have moved out of the cache.  If an older advertisement is asked for again, it will be regenerated. ...unless the content has been removed (in which case the indexer should not ask for it).

PR #168, opened 2022-02-02
Re implement caching mechanism as `CachedEntriesChunker`

### LINKED ISSUE #166: Entries chain cache grows indefinitely
During [chunk generation](https://github.com/filecoin-project/index-provider/blob/main/engine/linksystem.go#L138) if a single linked list is more than the cache's current capacity, the cache is resized to hold the entire linked list.

However the final resize is always to the length of the cache itself, which means none of the entries are ever evicted and cache grows indefinitely.

Test:
```golang
func Test_CacheIsEvicted(t *testing.T) {
	cfg := config.NewIngest()
	cfg.LinkedChunkSize = 10
	cfg.LinkCacheSize = 50
	engine := mkEngineWithConfig(t, cfg)

	mhCount := 100
	expectedChunkCount := mhCount / cfg.LinkedChunkSize

	for i := 0; i < 10; i++ {
		idx := index.NewMultihashSorted()
		cids, err := testutil.RandomCids(mhCount)
		require.NoError(t, err)
		var records []index.Record
		for i, c := range cids {
			records = append(records, index.Record{
				Cid:    c,
				Offset: uint64(i + 1),
			})
		}

		err = idx.Load(records)
		require.NoError(t, err)
		iterator, err := provider.CarMultihashIterator(idx)
		require.NoError(t, err)

		_, err = engine.generateChunks(iterator)
		require.NoError(t, err)
	}

	store, ok := engine.cache.(*lrustore.LRUStore)
	require.True(t, ok)
	require.Equal(t, expectedChunkCount, store.Cap())
}
```

masih:
Introduce the `EntriesChunker` interface that given a multihash iterator
would chunk the supplied multihashes into their interlinked `EntryChunk`
IPLD representation.

The chunker includes both chunking and caching mechanism because, the
cache must guarantee that either an entire chunk is cached or none of
it, while maintaining that guarantee for overlapping chains, i.e. chains
that share the same section.

Use the same datastore prefix as the old `lrustore` such that upon
instantiation the previously persisted values, if any, are pruned. Since
eviction did not function and any existing cache is likely larger than
capacity. In addition, the new implementation stores keys to the root
of chains and a counter for repeated links, which are required for the
`CachedEntriesChunker` to work correctly.

Fixes #166, #73


masih:
@gammazero Thank you for the discussion on slack. I am going to expand the scope of this PR to check that partially evicted entries from a chain don't cause any issues since before this fix cache always retained everything.

Keeping as draft until other tests confirm partially cached entry chains are handled gracefully.

engine/linksystem.go
```diff
@@ -188,7 +188,12 @@ func (e *Engine) generateChunks(mhIter provider.MultihashIterator) (ipld.Link, e
 	// If the cache was resized to expand beyond its original capacity, then
 	// set its size to only as big as the number of links in this list.
 	if resized {
-		ls.Resize(context.Background(), ls.Len())
+		newSize := e.linkCacheSize
+		if chunkCount > e.linkCacheSize {
+			newSize = chunkCount
+		}
```
gammazero, line 194: The line
```go
ls.Resize(context.Background(), ls.Len())
```
Was resizing the cache (`ls') down to the number of items current stored, because the rest of the capacity is not needed.

Your change sets the size back to the original size, which will have the effect of immediately removing chunks that were just created and put into the cache.

I don think we ever want to resize back down to the original size.  The cache size should be whatever is needed to hold the last generated set of chunks.

gammazero (changes requested):
This is not what we want to do here, see comments.

engine/linksystem.go
masih: > Was resizing the cache (ls) down to the number of items current stored, because the rest of the capacity is not needed.

That line always set the size to the number of items in the cache; which means none of the items in it ever get evicted and it grows indefinitely unless I am missing something. See https://github.com/filecoin-project/index-provider/pull/168/commits/c432f6dea45c704e34e6e601507207bd8881374f for local reproduction of cache growing as large as all the entries ever generated.

> The cache size should be whatever is needed to hold the last generated set of chunks.

Assuming the resizing we are looking for here should work as the comments added above the if statement, then the cache _capacity_ should either be as large as the configured capacity or as large as the longest generated chain which I think the code here does.


Let's discuss this further in a sync call.

masih:
@gammazero here is a test that asserts partially cached entries are handled gracefully and re-cached: https://github.com/filecoin-project/index-provider/pull/168/commits/6440af595e3c3df66fb30c681349902a1536930c


But the current implementation still has a race condition, where when the advertisements are synced in parallel, entries for a single advertisement may be re-generated many times and cached depending on the cache capacity.

engine/chunker/cached_chunker.go
```diff
@@ -0,0 +1,516 @@
...
+		}
+	} else {
+		log.Debugw("Cache restored successfully", "restoredCacheCount", ls.Len(), "capacity", ls.Cap())
+	}
+
+	// If the cache capacity was too small to restore all entries present, it means cache was
+	// evicted during restore and records were pruned as needed.
+	//
+	// Log an informative message to let the user know.
+	if ls.Cap() < count {
```
gammazero, line 379: Should this be moved into the above `else` block?  This does not seem useful if `count` is 0.

engine/chunker/cached_chunker.go
```diff
@@ -0,0 +1,516 @@
...
+	lCtx := ipld.LinkContext{Ctx: ctx}
+	next := root
+	for {
+		n, err := ls.lsys.Load(lCtx, next, schema.Type.EntryChunk)
+		if err != nil {
+			return links, err
+		}
+		chunk, ok := n.(schema.EntryChunk)
+		if !ok {
+			return links, fmt.Errorf("expected schema.EntryChunk for link: %s", next)
```
gammazero, line 415:
```suggestion
			return links, fmt.Errorf("expected schema.EntryChunk for link: %s", next.(cidlink.Link).Cid)
```
Do you need to convert the `next` link to a CID to get that string in the error message?

engine/chunker/cached_chunker.go
```diff
@@ -0,0 +1,516 @@
...
+	oVal, err := ls.ds.Get(ctx, oKey)
+	var count uint64
+	if err == datastore.ErrNotFound {
+		count = 1
+		oVal = make([]byte, 8)
+	} else if err != nil {
+		return err
+	} else {
+		count = binary.LittleEndian.Uint64(oVal) + 1
+	}
```
gammazero, line 477:
```suggestion
	if err != nil {
		if err == datastore.ErrNotFound {
			count = 1
			oVal = make([]byte, 8)
		} else {
			return err
		}
	} else {
		count = binary.LittleEndian.Uint64(oVal) + 1
	}
```

nit: eliminate a comparison when `err == nil`

gammazero (approved):
Looks great! Only a couple of trivial comments.

engine/chunker/cached_chunker.go
masih: I don't think we need to, because `ipld.Link`[ implements `Stringer` interface](https://github.com/ipld/go-ipld-prime/blob/master/datamodel/link.go#L33-L36).

The change suggested above:
1.  is equivalent to the code that is there already, since `cidlink.Link.String()` simply calls `cid.String()` [here](https://github.com/ipld/go-ipld-prime/blob/master/linking/cid/cidLink.go#L30-L32), and
2. assumes `next` is of type `cidlink.Link`; sure that is true today but if IPLD makes ca change to that for whatever reason we end up getting a panic in such lines. 


Try running this:
```go
func Test_StringerErrorMessage(t *testing.T) {
	c, err := cid.Decode("baguqeeqq4dalrhkuucrvzkmy6louau2wm4")
	if err != nil {
		t.Fatal(err)
	}
	l := cidlink.Link{Cid: c}

	errFromCid := fmt.Errorf("%s", c)
	errFromLink := fmt.Errorf("%s", l)

	if errFromCid.Error() != errFromLink.Error() {
		t.Fatal("andrew is right")
	}
}
```
 🙂

PR #170, opened 2022-02-08
Add option to configure miner ID for datatransfer publisher

gammazero:
If the datatransfer publisher is configured with a miner ID, then that is included with pubsub messages.  It is used to allow lotus chain-nodes to authenticate the message for relay.

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/filecoin-project/go-data-transfer v1.14.0
-	github.com/filecoin-project/go-legs v0.2.7
+	github.com/filecoin-project/go-legs v0.2.8
 	github.com/filecoin-project/index-provider v0.2.1
 	github.com/filecoin-project/storetheindex v0.3.0
 	github.com/ipfs/go-cid v0.1.0
--- a/engine/option.go
+++ b/engine/option.go
@@ -1,0 +1,32 @@
+package engine
+
+import "fmt"
+
+// engineConfig contains all options for engineConfiguring Engine.
+type engineConfig struct {
+	extraGossipData []byte
+}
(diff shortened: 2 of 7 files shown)


engine/option.go
```diff
@@ -0,0 +1,28 @@
...
+	for i, opt := range opts {
+		if err := opt(c); err != nil {
+			return fmt.Errorf("option %d failed: %s", i, err)
+		}
+	}
+	return nil
+}
+
+// WithMinerID sets the miner ID to include in the pubsub announcement.
+func WithMinerID(minerID string) Option {
```
willscott, line 23: even here i'd have index provider interface be neutral:
`WithExtraGossipData(extraData string) Option` or similar

cmd/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/filecoin-project/go-data-transfer v1.14.0
-	github.com/filecoin-project/go-legs v0.2.7
+	github.com/filecoin-project/go-legs v0.2.8-0.20220208004129-dc48d0d70036
```
masih, line 7: Is the plan to depend on a release of go-legs before this PR is merged?

willscott: yeah, once we merge the legs PR we'll want to cut a release and rebase

engine/option.go
gammazero: OK, hang on...

gammazero: Prefer `string` or `[]byte`?

masih: I would say whatever it is in legs

masih (approved):

masih:
Merging since macOS CI is taking a long time. Tests on other platforms are green.

PR #173, opened 2022-02-10
Bump version to include go-legs message format change and bug fixes

masih:
Bump version prior to release, in order to include two main changes into
SPX rollout:
 1. go-legs message format change, and
 2. caching eviction bug fix.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,3 +1,3 @@
 {
-  "version": "v0.2.7"
+  "version": "v0.3.0"
 }


version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "v0.2.7"
+  "version": "v0.2.8"
```
willscott, line 2: may need to be v0.3.0 since protocol change

willscott (approved):

PR #176, opened 2022-02-11
Delay republishing last advertisement notification on start

gammazero:
The delay is needed to give some time to connect to othernodes. Otherwise, if not connected to other nodes, the notification will never go out.

An alternative to this PR is to have any applications, that use this provider, call `engine.PublishLatest` sometime after connecting to other node(s).

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -1,11 +1,14 @@
 package config
 
+import "time"
+
 const (
 	// Keep 1024 chunks in cache; keeps 256MiB if chunks are 0.25MiB.
 	defaultLinkCacheSize = 1024
 	// Multihashes are 128 bytes so 16384 results in 0.25MiB chunk when full.
-	defaultLinkedChunkSize = 16384
-	defaultPubSubTopic     = "/indexer/ingest/mainnet"
+	defaultLinkedChunkSize   = 16384
+	defaultPubSubTopic       = "/indexer/ingest/mainnet"
+	defaultStartPublishDelay = Duration(time.Minute)
 )
 
 type PublisherKind string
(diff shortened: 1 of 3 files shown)


masih:
I don't think a static delay is a good idea. At the very least we should make that delay configurable. 

I wonder if it makes sense to expand the markets api to add a CLI command that lets the markets process republish head. 

Going to deploy this anyway to check gossip behaviour

gammazero:
@masih The delay is now configurable.

config/ingest.go
```diff
@@ -36,16 +39,23 @@ type Ingest struct {
 
 	// PublisherKind specifies which legs.Publisher implementation to use.
 	PublisherKind PublisherKind
+
+	// StartPublishDelay specifies how long to wait, after startup, to
+	// re-publish the latest advertisement notification.  This delay gives time
+	// to connect to other nodes and establish gossip mesh before sending
+	// pubsub advertisement notification message.
+	StartPublishDelay Duration
```
masih, line 47: `PublishLatestAdOnStartDelay`? Or a name that makes it clear this only applies to republishing of the latest ad?

masih (approved):

config/ingest.go
gammazero: I think I would rather change this to:
```go
// AutoAnnounceInterval is the amount of time to elapse, in which no new announcements
// have been published, until the latest advertisement CID announcement is republished. 
AutoAnnounceInterval Duration
```
and default it to 12 hours, and have a command `provider announce` to announce the latest ad CID on demand.

gammazero:
Closing this in favor of having an `announce` command to optionally publish an announcement. See #205

PR #183, opened 2022-02-23
Upgrade indexing dependencies and bump version

masih:
Upgrade dependency to `storetheindex` and `go-legs`.

Bump version in preparation for release.


willscott (approved):

masih:
Build fails due to removal of an option to supply datatrasfer to a legs subscriber. 
I can hack at the tests to get them running since the DT option is removed from legs. but I think we should have that option back in legs. Captured https://github.com/filecoin-project/go-legs/issues/95

PR #184, opened 2022-02-24
Use legs subscriber to abstract out the dtsync vs httpsync

MarcoPolo:
...In list.

I was testing this on an http endpoint and it was failing. I simplified the two flows into a single flow using the legs subscriber abstraction.

I also fixed a bug that prevented entries from showing in the list command.

### DIFF
--- a/cmd/provider/internal/client_store.go
+++ b/cmd/provider/internal/client_store.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ipfs/go-datastore"
 	dssync "github.com/ipfs/go-datastore/sync"
 	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/linking"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/multicodec"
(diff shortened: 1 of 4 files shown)


cmd/provider/internal/client_store.go
```diff
@@ -69,22 +70,12 @@ func newProviderClientStore() *ProviderClientStore {
...
-		return cid.Undef, nil, err
-	}
-	nb := schema.Type.EntryChunk.NewBuilder()
-	decoder, err := multicodec.LookupDecoder(target.Prefix().Codec)
-	if err != nil {
-		return cid.Undef, nil, err
-	}
-
-	err = decoder(nb, bytes.NewBuffer(val))
+	n, err := s.LinkSystem.Load(linking.LinkContext{}, cidlink.Link{Cid: target}, schema.Type.EntryChunk)
```
masih, line 73:
```suggestion
	n, err := s.LinkSystem.Load(linking.LinkContext{Ctx: ctx}, cidlink.Link{Cid: target}, schema.Type.EntryChunk)
```

masih (approved):
Nice one thanks @MarcoPolo 🍻

PR #187, opened 2022-03-03
Use concrete multicodec codes as indexing metadata protocol ID

masih:
Refactor metadata package to: 1) use concrete multicodec codes defined
in the lookup table, and 2) simplify the datatransfer metadata
implementation since dynamic protocol ID range no longer makes sense
when using concrete multicodec codes.

Relates to:
 - https://github.com/multiformats/multicodec/pull/260


metadata/graphsync_filecoinv1.go
```diff
@@ -0,0 +1,87 @@
...
+	nd := bindnode.Wrap(dtm, graphSyncfilecoinV1SchemaType)
+	buf := new(bytes.Buffer)
+	err := dagcbor.Encode(nd, buf)
+	if err != nil {
+		return stiapi.Metadata{}, err
+	}
+	return stiapi.Metadata{
+		ProtocolID: multicodec.TransportGraphsyncFilecoinv1,
+		Data:       buf.Bytes(),
+	}, nil
```
willscott, line 54: Just a note, but i think we can in a protocol-compatible way, say that metadata is not simply `<protocol ID><payload>` but rather `[<protocol id><protocol-speicfic-payload>]`, so e.g.
`uvarint{graphsync filecoin} || cbor{<cid>, true, true} || uvarint{bitswap}` to indicate that both protocols are supported.
since the cbor payload of graphsync can already be determined where the end is, if there are trailing bytes they can be interpreted as the next protocol

willscott (approved):

metadata/graphsync_filecoinv1.go
masih: Captured: https://github.com/filecoin-project/index-provider/issues/188 
Leaving this out for another PR.


If I understand the comment correctly, it means the protocol ID GraphsyncFileCoinV1 can specify multiple retrieval protocols?  If so this might seem slightly odd, where putting Bitswap metadata first means the user can't specify multiple.  Is the suggestion that, basically metadata can be nested?

From API perspective I am not sure if this needs changes in the go-indexer-core.

Cc @gammazero

willscott: Nope, it would be in either order. Probably we'd say that when multiple protocols are supported by the provider they should be listed in metadata in numerical order (so bitswap, then data transfer).

The constraint i'm noting we need for that is that payloads need to be parseable in a way where we know when they've ended. We're lucky in that our defined payloads so far do have that property all ready, so this is a backwards compatible change to the definition of metadata structure.

masih: Gotcha thank you

Issue #82, opened 2021-11-04
Use IPLD schema generator to defile `FilecoinV1Data` metadata schema


masih:
The [metadata schema](https://github.com/filecoin-project/indexer-reference-provider/blob/39604c8839e2f0ac8c6a1e097b1ead8e272596f3/metadata/metadata.go#L82-L94) for `FilecoinV1Data` metadata is defined programatically. 

Now that the schema generator work in IPLD is operational, use the generator similar to the work done in [storetheindex](https://github.com/filecoin-project/storetheindex/pull/102).

masih:
Use bind node instead since codegen is no longer recommended.

PR #201, opened 2022-03-17
Replace IPLD accumulate with IPLD schema file

masih:
Add a dedicated IPLD schema type for GraphSyncFilecoinV1 metadata
protocol to replace the schema created using IPLD `Accumulate`.

Fixes: #82


mvdan (approved):
:ok_hand:

PR #224, opened 2022-05-10
Do not assume build system has bash shell

### LINKED ISSUE #223: Set  flags based on shell
Either add the following above where .SHELLFLAGS is set:
SHELL = bash
Or, assume only standard bourne shell and do not set bash-specific flags.

See: https://github.com/filecoin-project/index-provider/commit/b51ec719920fac17aa7f1e517c8bf3b5c5650d86#r73043634

gammazero:
- The build system may not have bash shell installed, so do not use bash-specific shell flags.  Instead only assume standard bourne shell capabilities.
- Do not do `git fetch --tags` since it is usually not necessary and requires github access or explicitly setting `TAG`.
- Fix .gitignore to correctly ignore provider executable.

Fixes #223

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-./provider
+provider
 dist/
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
-.SHELLFLAGS := -eu -o pipefail -c
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
-TAG ?= $(shell git fetch --tags && git describe --tags --abbrev=0 | sed 's/v//')
+TAG ?= $(git describe --tags --abbrev=0 | sed 's/v//')
 COMMIT ?= $(shell git rev-parse --short HEAD)
 CLEAN ?= $(shell git diff --quiet --exit-code || echo '-SNAPSHOT')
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,2 @@
-./provider
+provider
```
masih, line 1: @gammazero this will ignore any _directory_ or file named `provider` recursively, which will include any untracked files added to [`cmd/provider`](https://github.com/filecoin-project/index-provider/tree/main/cmd/provider), hence the change to add `./` and make it explicit that the line here is meant to ignore the built binary.

=

Makefile
```diff
@@ -1,8 +1,7 @@
-.SHELLFLAGS := -eu -o pipefail -c
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
-TAG ?= $(shell git fetch --tags && git describe --tags --abbrev=0 | sed 's/v//')
```
masih, line 5: Missing `shell` after `$(`

gammazero: Doh!  Replaced the shell.

I do have my config set up to always do ssh.  Anyway, as you point out, make targets are used purely as utility by dev...  so I'll just put that back as well.

PR #226, opened 2022-05-10
supporting publish custom advertisement

zl03jsj:
Hello! I am one of the contributors to Venus Filecoin and have been working on the integration of venus-market with index-provider.

We propose the following approach for the integration: 

We want minimize the coupling between index-provider and venus-market.

The main idea is, index-provider expose an http api, something like 'publishAdvertisement'.
venus-market collect changed indices, and send to index-provider through this 'publishAdvertisement', for publishing.
We think it's a good approach to decoupling.
The venus-market even doesn't need to import 'index-provider' as dependent package.

request param is defined as following:
```go
AnnounceAdvReq struct {
        // full index pairs like:{cid: offset}, index-provider convert it into multi-hash iterator
	Indices         []provider.IteratorStep 
	ContextID       []byte // deal proposal cid
	ProviderID      peer.ID 
	RetrievalAddrs  []string
	Md              metadata.Metadata
	IsDel           bool // if isDel is true, `Indices` is ignored.
}
// the array size of  `Indices` is 131072, for a 32G file. each pair size is 12 bytes, 
// its total byte size small than 2M, it is acceptable for http requests.
```

to supporting out calls, it's necessary to expose `Engine.publishAdvForIndex` as public,with more params:
```go
PublishAdvForIndex(ctx context.Context, contextID []byte, mhIter provider.MultihashIterator,
	providerID peer.ID, retrievalAddrs []string, md metadata.Metadata, isRm bool) (cid.Cid, error)
```

Please review and let us know what you guys think of this approach. Thank you!


server/admin/http/models.go
```diff
@@ -53,3 +56,20 @@ type (
 		Paths []string `json:"paths"`
 	}
 )
+
+type (
+	// AnnounceAdvReq represents a request for publishing a raw advertisement.
+	AnnounceAdvReq struct {
+		Indices         []provider.IteratorStep
```
gammazero, line 63: If this is meant to be passed over an HTTP interface, I think `Indices` would be better as type `[]multihash.Multihash`, since there is no need for the offsets. The slice of `multihash.Multihash` should be correctly ordered by the client.

Also, since you do not need the offsets, there is no need to reuse the existing `MultihashIterator` implementation `indexMhIterator`.  Better to create your own `MultihashIterator` implementation.  Maybe it could look something like:
```go
type indexMhIterator struct {                                                                                                                
    mhs []multihash.Multihash                                                                                                                
    pos int                                                                                                                                  
}                                                                                                                                            
func (i *indexMhIterator) Next() (multihash.Multihash, error) {                                                                              
    // We have reached the end of stream.                                                                                                    
    if i.pos >= len(i.mhs) {                                                                                                                 
        return nil, io.EOF                                                                                                                   
    }                                                                                                                                        
    mh := i.mhs[i.pos]                                                                                                                       
    i.pos++                                                                                                                                  
    return mh, nil                                                                                                                           
}                                                                                                                                            
```
This would be constructed in ` publishAdvHandler()` :
```go
mhIter = indexMhIterator{
    mhs: req.Indices,
}
```

index_mh_iter.go
```diff
@@ -12,31 +12,35 @@ import (
...
 	curStep    int
 	lastOffset uint64
 }
 
-type iteratorStep struct {
-	mh     multihash.Multihash
-	offset uint64
+type IteratorStep struct {
+	Mh     multihash.Multihash
+	Offset uint64
```
gammazero, line 23: Do not reuse this implementation if you do not need to have offsets.

.gitignore
```diff
@@ -1,2 +1,3 @@
 ./provider
 dist/
+provider
```
gammazero, line 3: Please remove this change from the PR.  It has already been handled in #227.

gammazero (changes requested):

server/admin/http/models.go
zl03jsj: make sense. i will fix this.

zl03jsj: > Thanks for the contribution @zl03jsj !
> 
> I wonder if there's a way to do this that would help us have a bit less additional code:
> 
> * you introduce a new struct type, `AnnounceAdvReq` on the admin server for data
> * This looks a lot like what lotus uses in its library interface to [`Publish`](https://github.com/filecoin-project/index-provider/blob/main/engine/engine.go#L191) - do you think we could re-use that type and method as what the handler calls?

i think it's worth a try. and after some researching on this, 
if we re-use  the `Engine.Publish` method , it is have to create `schema.Advertisement` out side the `engine`, 
```go
	Advertisement struct {
		PreviousID *ipld.Link
		Provider   string
		Addresses  []string
		Signature  []byte
		Entries    ipld.Link
		ContextID  []byte
		Metadata   []byte
		IsRm       bool
	}
```
to fill the its members, such as:`PreviousID`, `Entries`, `Signature`, 
the `Engine` have to expose some **protected** members outside(e.g:`key`, `entriesChunker`..), 
https://github.com/filecoin-project/index-provider/blob/cf3e8e80afdc62bf0ce94cfcb375f73e751b50cb/engine/engine_api_test.go#L19-L26

they are now only for test purpose. do your think is that ok?

and on the other side,  in the old approach, 
we just needs minor change on `PublishAdvForIndex` and make it public is ok.

i prefer expose `PublishAdvForIndex`.  (:-

zl03jsj:
@gammazero needs a review,  fixes based on your advice.

engine/engine.go
```diff
@@ -256,7 +258,7 @@ func (e *Engine) RegisterMultihashLister(mhl provider.MultihashLister) {
 func (e *Engine) NotifyPut(ctx context.Context, contextID []byte, md metadata.Metadata) (cid.Cid, error) {
 	// The multihash lister must have been registered for the linkSystem to know how to
 	// go from contextID to list of CIDs.
-	return e.publishAdvForIndex(ctx, contextID, md, false)
+	return e.PublishAdvForIndex(ctx, contextID, nil, e.provider.ID, e.retrievalAddrsAsString(), md, false)
```
masih, line 261: Further to comments from @willscott please consider using the existing `Publish` API that takes `schema.Advertisement` which closely matches the HTTP request struct introduced in this PR and avoid exposing internal Engine APIs :)

server/admin/http/server.go
```diff
@@ -48,6 +48,10 @@ func New(h host.Host, e *engine.Engine, cs *supplier.CarSupplier, o ...Option) (
 		Methods(http.MethodPost).
 		Headers("Content-Type", "application/json")
 
+	r.HandleFunc("/admin/publish_adv", s.publishAdvHandler).
```
masih, line 51: I am not well rehearsed on the Venus Market use case; I recommend thinking carefully before adding this to the `admin` HTTP server. That server is meant to _never_ be exposed publicly and should only be accessible locally with the host running the standalone index provider daemon. If this endpoint is to be exposed, consider either:
* implementing your own HTTP server that wraps the existing exported Engine APOIs for publishing advertisements, or 
* Introduce an `engine.Option` that instantiates a public facing HTTP endpoint for publishing advertisements separate from the admin HTTP server.

I'd also recommend using a simpler path for this handler, e.g. `/ad` or `/advertisement` that accepts `POST`.

masih (changes requested):

server/admin/http/models.go
masih: @zl03jsj the rationale behind having the entires chunks in HTTP terms is simply pagination. The HTTP endpoint wherever it ends up at will need some form of pagination for the multihash entries since they can certainly be very long. 

I recommend thinking about what the pagination substitute would be for the HTTP endpoint.

server/admin/http/server.go
zl03jsj: your mention is helpful, i will think about this.

engine/engine.go
zl03jsj: since publish a `advertisement` from http request is with the same logical of `publishAdvForIndex`,
we needs to manage the `metadata` of advertisements,  
so following methods in `Engine` are all needs by this logical:

https://github.com/filecoin-project/index-provider/blob/cf3e8e80afdc62bf0ce94cfcb375f73e751b50cb/engine/engine.go#L455-L510

i had successfully construct an `Advertisement` out `Engine`, and call `Publish`, this is ok.

but if do **NOT** expose any method,  you known, it is hard. no way to manage the `metadata`

i still think the good way is expose `publishAdvForIndex`.

hope your suggestion. thanks..

masih: > no way to manage the metadata

Can you expand on this please.

zl03jsj: ok, i may find another way to do this.
not expose `publishAdvForIndex` and have no change to `Engine`. will working on this.. (:-

zl03jsj: > > no way to manage the metadata
> 
> Can you expand on this please.

tomorrow, it's getting late in china now.

zl03jsj: > Can you expand on this please.

ok, i will try to explain the `hard`, 
assume we create a advertisement http server to handle advertisement publishing request,  and not expose any method in `engine`, 

this http server should be defined like this:
```go
Server struct {
      ....
	e  *engine.Engine
	k  crypto.PrivKey
	ce *chunker.CachedEntriesChunker //
	ds datastore.Batching
}
```

just a **not remove** advertisement for example, the logical of handler should be like:
```go
func (s *Server) advertisementHandler(w http.ResponseWriter, r *http.Request) {
	var (
		req AnnounceAdvReq
		err error
	)
       ......
	ctx := r.Context()
	
	contextCID, _ := s.ds.Get(cidMapPrefix+req.ContextID)
	if contextCID == cid.Undef {
		lnk := ... generate link list ...
		s.ds.Put(ctx, cidMapPrefxi+req.ContextID, lnk.Cid)
	} else {
		prevMetadata, _ := e.ds.Get(keyMetadataMapPrefix+req.ContextID)
		if req.Md.Equal(prevMetadata) {
			... response error ...
		}
	}
	s.ds.Put(ctx, keyToMetadataMapPrefix+req.ContextID, &req.Md)

	adv, err := s.makeAdvertisement(ctx, ...)
	if err != nil {
		... resposne error ...
	}
	adv.Sign(s.k)
	
	// finally we can call `engine.Publish`
	s.e.Publish(ctx, adv)
	
	... respond result ...
}
```

what the `Server.advertisementHandler` do is an exactly copy of `Engine.PublishAdvForIndex`.

i said hard to mange `metadata`, means those  `e.ds.Get`, `e.ds.Put` in above code,  are another exactly copy of operations of the badger store , 
witch is already implemented in `Engine`

steven004:
If I understand correctly, @zl03jsj is willing to run Index Provider as a standalone service which can provide more flexibilities. I would think it is great and it's a reasonable requirement. 

Thus, there will be more parameters to be transferred from Market service to Index-Provider since they are not sharing datastore. We need to find a way, basically via an API to pass more information, such as providerID, retrivalAddress. 

Adding or exposing an interface is one way, or we might change the notify to accommodate more data? or any other suggestions? @masih @willscott

engine/engine.go
masih: Re Context ID and metadata determination in this code snippet:
```go
contextCID, _ := s.ds.Get(cidMapPrefix+req.ContextID)
	if contextCID == cid.Undef {
		lnk := ... generate link list ...
		s.ds.Put(ctx, cidMapPrefxi+req.ContextID, lnk.Cid)
	} else {
		prevMetadata, _ := e.ds.Get(keyMetadataMapPrefix+req.ContextID)
		if req.Md.Equal(prevMetadata) {
			... response error ...
		}
	}
	s.ds.Put(ctx, keyToMetadataMapPrefix+req.ContextID, &req.Md)
```

This is _entirely application specific_. The point of libraries offered by this repository is to be application agnostic. That logic cannot live in this repo. Because, the way Context ID or Metadata determined by different application is different. And that's why they are of type `[]byte` in the advertisement schema.

An alternative approach to make the implementation of the wrapper `Server` on your side simpler is to use `provider.MultihashLister` and [register](https://github.com/filecoin-project/index-provider/blob/main/interface.go#L40) an implementation of it to the `Engine`. That would then allow the logic by which a `contextID` can retrieve the list of multihashes for a an advertisement to be customised. The server can then use `NotifyPut` and `NotifyRemove` to publish advertisements instead of calling `Publish`.

willscott:
It sounds like what @masih is suggesting is that there can be an additional repo that holds the 'filecoin-index-provider-server' service wrapper around this library.

it sounds like the use-case-specific customization won't be a huge amount compared to functionality in the core library here, but the use case is such that it may be worth a separation of concerns, as has been done with e.g. IPFS datastore as an index provider.

masih:
To add to what @willscott pointed out: I am happy for the additional functionality to live in this repo as long as:
* it isolated in terms of code structure
* it is optional, e.g. explicitly enabled via some `engine.Option` or in a separate package with clear godocs explaining usage etc.

engine/engine.go
zl03jsj: This is a good idea
I had tried this solution yesterday, the reason i gave up was i didn't kwon how to pass `ProviderID`, `Retrieval address` to the `Engine`.
and you just said:
> ProvideID can be passed today to the Engine by specifying libp2p [host](https://github.com/filecoin-project/index-provider/blob/main/engine/options.go#L224). Retrieval address can also be specified by [this](https://github.com/filecoin-project/index-provider/blob/main/engine/options.go#L245) option

I m a little confused about this,  when the `daemon` create the `engine` instance:
https://github.com/filecoin-project/index-provider/blob/cf3e8e80afdc62bf0ce94cfcb375f73e751b50cb/cmd/provider/daemon.go#L119-L126
the `host` of index provider is created here, according to your above suggestion, does it means:
I should  create another **provider** `host`,  may be need add a configuration for this? 
like the following fake code:

```go
var providerHost = h
if provider_configuration_exists {
   providerHot = createProviderHost()
}
// Starting provider core
eng, err := engine.New(
...
	engine.WithHost(providerHost),
...

```

thanks for your patience @masih

masih: No worries; happy to be of help.

The `daemon` is meant to be a service that publishes CAR files only. A bit of history: it was written as a tool for internal testing when network indexer was being developed. 

As it stands today, `daemon` reads that information from a config file. So you can customise those values in the config file, and set any options not set from the config when `daemon` is instantiated. But keep in mind that `daemon` is not designed for dynamically changing retrieval address etc.

A better approach here may be to implement a separate HTTP server that takes an `engine` instance, then optionally expose that via `daemon` as a dedicated server. If that code is going to be application specific, then my recommendation is to move that to a repo close to where Venus market is located, and try not to reuse `daemon`. Instead start up that wrapper directly in a `main` function.

zl03jsj:
> @steven004 `ProvideID` can be passed today to the Engine by specifying libp2p [host](https://github.com/filecoin-project/index-provider/blob/main/engine/options.go#L224). Retrieval address can also be specified by [this](https://github.com/filecoin-project/index-provider/blob/main/engine/options.go#L245) option. Can you enumerate any other parameters that need to be customised? That should help me understand what it is that needs to be customised and is not currently possible via the API.

But there till have some problems:
If there are multiple `data-provider` send advertisement http requests to this `http-index-provider`,
The `ProviderID` and `RetrievalAddress` of advertisements are different, the above approach is not applied.

@masih @steven004

engine/engine.go
zl03jsj: Tanks a lot. I think i get it. .....LOL...
I should not think of the index-provider to be an app, it should be a library. and the cmd/provider is just a testing tool.
I had been getting stuck in the thinking of adding this functionary on this testing tool.

hunjixin:
@willscott 
I noticed that there is such an interface in the storethehash project [importManifest](https://github.com/filecoin-project/storetheindex/blob/0298fe0d9c67b1981d34c1aa37299c66b49cf7dd/server/admin/http/handler.go#L165) 
this interface is more needed. At present, the two problems with this interface are:
1. This interface is the admin interface
2. This api needs to get the result by reading the a file.

I know this is probably a test api, but this api does what we need, and it's really easy to use.

PR #228, opened 2022-05-10
Smaller default MultihashIterator implementation

gammazero:
The default `MultihashIterator` implementation does not need to keep the offsets if a check is done for duplicates when the iterator is created. This reduces memory allocated for each iterator instance, by n uint64 values, at the cost of one O(n) iteration of the sorted offsets to check for duplicates and copying n multihash values.

The default implementation is also exported so that it can be used in other packages that only need a simple multihash iterator.

To reviewers: This may only be worth it if there are other packages that end up reusing this iterator implementation, such as in #226


willscott (approved):

index_mh_iter.go
```diff
@@ -9,13 +9,13 @@ import (
...
+var _ MultihashIterator = (*MhIterator)(nil)
 
-type indexMhIterator struct {
-	steps []iteratorStep
-
-	curStep    int
-	lastOffset uint64
+// MhIterator is a simple MultihashIterator implementation that iterates a
+// slice of multihash.Multihash.
+type MhIterator struct {
```
masih, line 16: * I am curious why the PR exports this type? 
* Would it make sense to name it `SliceMultihashIterator` or something along those lines please? Having both `MultihashIterator` and `MhIterator` could be confusing to the user of the API.
  * We can then replace other utilitarian implementations like [this](https://github.com/filecoin-project/index-provider/blob/main/engine/slicemhiter_test.go).

gammazero: It was exported for _potential_ future use like in #226, so maybe best left unexported here and make that change in such future PR.

It would be better to call it `sliceMultihashIterator`.

PR #235, opened 2022-06-17
Implement the ability to publish ads with HAMT entries

masih:
Introduce the ability to configure index provider engine to publish
advertisements with entries structured as a HAMT with configurable
hash algorithm, bit-width and bucket size.

Refactor existing cached chunker as a generic DAG cache, and separate
the DAG generation logic into two implementation of non-cached chunkers:
one for `EntryChunk` chain and one for HAMT.

Refactor tests to assert that chunking mechanism for `EntriesChunk` work
as before and caching logic functions regardless of underlying DAG data
structure.


engine/chunker/cached_chunker_test.go
```diff
@@ -40,6 +41,10 @@ func TestCachedEntriesChunker_OverlappingLinkCounter(t *testing.T) {
 	require.Len(t, c1Chain, 2)
 	requireChunkIsCached(t, subject, c1Chain...)
 
+	for _, link := range c1Chain {
+		fmt.Println("> ", link.(cidlink.Link).Cid.String())
+	}
+
```
willscott, line 47: leftover debugging?

engine/chunker/hamt_chunker.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if err := ma.Finish(); err != nil {
+		return nil, err
+	}
+	hamtRoot, ok := builder.Build().(ipld.ADL)
+	if !ok {
+		return nil, errors.New("HAMT must implement ipld.ADL")
+	}
+	return h.ls.Store(ipld.LinkContext{Ctx: ctx}, schema.Linkproto, hamtRoot.Substrate())
+}
```
willscott, line 90: this is going to build the whole thing up in memory. i don't think there's a way to stream it, right?

masih: Not quite. Depends on the parameters and size of the tree. 
Basically when there are no more buckets left it will start adding links which are stored in the linksystem as the HAMT grows.

masih: No full streaming API in IPLD HAMT yet I'm afraid.

willscott (approved):

multihash_iterator.go
```diff
@@ -5,10 +5,15 @@ import (
...
 	"sort"
 
+	"github.com/filecoin-project/storetheindex/api/v0/ingest/schema"
 	carindex "github.com/ipld/go-car/v2/index"
+	hamt "github.com/ipld/go-ipld-adl-hamt"
+	"github.com/ipld/go-ipld-prime"
 	"github.com/multiformats/go-multihash"
 )
 
+var _ MultihashIterator = (*sliceMhIterator)(nil)
```
gammazero, line 15: nit: This is not necessary since `SliceMultihashIterator` returns `&sliceMhIterator` as a `MultihashIterator` interface, thereby guaranteeing that `sliceMhIterator` implements `MultihashIterator`.

This check is only needed when the constructor returns the concrete type or a different interface type from the target interface.

Maybe better to return the concrete type, and let the calling code pass it around as a `MultihashIterator`?

multihash_iterator.go
```diff
@@ -66,3 +70,33 @@ func (it *sliceMhIterator) Next() (multihash.Multihash, error) {
 	it.pos++
 	return mh, nil
 }
+
+var _ MultihashIterator = (*ipldMapMhIter)(nil)
```
gammazero, line 74: nit: Not necessary, because `HamtMultihashIterator` returns `&ipldMapMhIter` as `MultihashIterator` thereby guaranteeing that the interface is implemented.

engine/chunker/hamt_chunker.go
```diff
@@ -0,0 +1,101 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if err := ma.AssembleKey().AssignBytes(mh); err != nil {
+			return nil, err
+		}
+		if err := ma.AssembleValue().AssignBool(true); err != nil {
```
gammazero, line 89: I am assuming `Bool` is the smallest value type. Is there something smaller, like a nil type that would also make it clear that the value is not being used for anything?

gammazero (approved):
I do not see any absolutely necessary changes, but did have a few comments and questions.

- Should there be a config item to enable HAMT entries and to set the hash algo?
- What happens on the indexer if separate ad publishers choose different HAMT algos, but are indexing the exact same data for the same provider? What effect will that have on caching and archiving entry content?

engine/chunker/chain_chunker.go
```diff
@@ -0,0 +1,101 @@
...
+	var next ipld.Link
+	var mhCount, chunkCount int
+	for {
+		mh, err := mhi.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
masih: I'm curious do you find this style clearer to read? The rationale for writing it the way I did is to have one less extra check for EOF case since `nil` will not equal `EOF`

multihash_iterator.go
masih: This is more as a documentation that the structs should implement `MultihashIterator`

engine/chunker/hamt_chunker.go
masih: Not sure. Happy to iterate on this. The implementation here follows the [docs](https://ipld.io/specs/advanced-data-layouts/hamt/spec/#use-as-a-set) on using HAMT as set.

PR #236, opened 2022-06-17
Add command to publish announce message via HTTP

gammazero:
The `announce-http` command manually announces the latest advertisement CID to a specific indexer over HTTP. This works whether the provider publishes advertisements over dtsync or httpsync.

### DIFF
--- a/cmd/provider/announce.go
+++ b/cmd/provider/announce.go
@@ -3,8 +3,10 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"net/http"
 
+	adminserver "github.com/filecoin-project/index-provider/server/admin/http"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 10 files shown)


cmd/provider/announce.go
```diff
@@ -32,3 +43,36 @@ func announceCommand(cctx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return errFromHttpResp(resp)
+	}
+
+	_, err = cctx.App.Writer.Write([]byte("Announced latest advertisement via HTTP\n"))
```
willscott, line 76: maybe include the cid or announcement serialization?

engine/host_to_maddr.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return ""
+}
+
+func hostToMultiaddr(hostport string) (multiaddr.Multiaddr, error) {
+	host, port, err := net.SplitHostPort(hostport)
+	if err != nil {
+		port = "80"
+		host = hostport
+	}
```
willscott, line 34: aren't there some forms of hostport that should actually error here?

gammazero: Errors should all be caught by `parseAddrType` and `multiaddr.NewMultiaddr`.  I added a unit test that checks failures.

cmd/provider/announce.go
gammazero: Responds with announced CID.

willscott (approved):

PR #245, opened 2022-06-21
Format comments

gammazero:
Format and improve some wording.


masih:
@gammazero  what max screen width do you use for formatting docs? I am happy to change mine to your setting to reduce whitespace changes.

Issue #246, opened 2022-06-24
Deploy an instance of standalone index-provider in `dev` environment for testing purposes


masih:
Deploy the standalone index-provider in the `dev` environment of storetheindex, so that we have a provider with full access to test changes in protocol and build confidence without having to disrupt other providers.

As the first use case, the instance will be used to test the HAMT ad publication

PR #247, opened 2022-06-27
Publish container image to ECR on merge to `main` and release

masih:
Add CI job to build and publish the container image to ECR on merges to
`main` and on release with deterministic tagging.

Add a CI job to build the dockerfile and make sure it is functional on
push and PR.

Fix a build issue in dockerfile, and remove the error-prone `init` build
argument. Because, a production environment would not want the docker
image to contain peer identity since it is typically mounted, and a
standalone testing app would benefit from a utility script to set things
up since as it stood the init in dockerfile was broken anyway.

Relates to: #246


Dockerfile
```diff
@@ -5,14 +5,9 @@ COPY go.mod go.sum ./
 RUN go get -d -v ./...
 
 ADD . /go/src/provider
-RUN go build -o /go/bin/provider
-
-# TODO consider auto initialization flag as part of `daemon` command
-ARG INIT_PROVIDER='true'
-RUN if test "${INIT_PROVIDER}" = 'false'; then /go/bin/provider init; else echo 'skipping provider initialization.'; fi
+RUN CGO_ENABLED=0 go build -o /go/bin/provider ./cmd/provider
```
willscott, line 8: there are no deps that need cgo right now?

willscott (approved):

PR #252, opened 2022-07-07
Send HTTP announce to URLs configured to receive it

### LINKED ISSUE #250: Add option to announce ads to an explicit indexer
Add an option that allows the provider engine to announce ads to an explicit indexer

gammazero:
- DirectAnnounce config contains indexer URLs to receive direct HTTP announce message
- When publishing a new advertisement send HTTP announce all configured indexer URLs
- Engine has WithDirectAnnounce option to set HTTP announce URLs
- Update dependencies

Fixes #250


engine/engine.go
```diff
@@ -250,43 +257,69 @@ func (e *Engine) PublishLatest(ctx context.Context) (cid.Cid, error) {
...
 		ID: e.h.ID(),
 	}
 
 	switch e.pubKind {
 	case NoPublisher:
 		log.Info("Remote announcements disabled")
-		return cid.Undef, nil
+		return nil
 	case DataTransferPublisher:
 		ai.Addrs = e.h.Addrs()
```
masih, line 286: For libp2p you can get the host's addrinfo directly from host via: `h.Peerstore().PeerInfo(h.ID())`.

engine/engine.go
```diff
@@ -250,43 +257,69 @@ func (e *Engine) PublishLatest(ctx context.Context) (cid.Cid, error) {
...
 		proto, _ := multiaddr.NewMultiaddr("/http")
 		ai.Addrs = append(ai.Addrs, multiaddr.Join(maddr, proto))
 	}
 
-	cl, err := httpclient.New(indexerHost)
-	if err != nil {
-		return cid.Undef, err
+	errChan := make(chan error)
+	for _, u := range announceURLs {
+		// Send HTTP announce to indexers concurrently.
```
masih, line 298: Do we need to close `errChan` somewhere?

engine/engine.go
```diff
@@ -250,43 +257,69 @@ func (e *Engine) PublishLatest(ctx context.Context) (cid.Cid, error) {
...
+				return
+			}
+			errChan <- nil
+		}(u)
 	}
 
-	if err = cl.Announce(ctx, ai, adCid); err != nil {
-		return cid.Undef, err
+	var errs error
+	for i := 0; i < len(announceURLs); i++ {
```
masih, line 316: `select` on `errChan` and `ctx`? We have context in this function already. Would be good to use it in case context has a timeout.

engine/options.go
```diff
@@ -43,6 +44,11 @@ type (
 	options struct {
 		ds datastore.Batching
 		h  host.Host
+
+		// announceURLs is the list of indexer URLs to send direct HTTP
+		// announce messages to.
+		announceURLs []*url.URL
```
masih, line 50: Thinking out loud: It's a bit strange that there is straight up URL, multiaddr for DT, _and_ our own flavour of HTTP multiaddr.

masih (changes requested):
One blocker on `select`ing `ctx`while waiting on `errChan` in `PublishLatestHTTP`. Otherwise, LGTM.

engine/engine.go
gammazero: No, all announce goroutines put a response onto `errChan`, and the look that collects the response gets a response for each goroutine, without needing to wait for the channel to close.  Once a response for each announce goroutine has been received, then all goroutines have completed.

engine/engine.go
gammazero: A select is not needed because `ctx` is passed into `cl.Announce`.  If `ctx` is canceled then all the pending requests will be canceled and return the canceled error.  That will cause the `errChan` reader to collect those errors and return.

engine/options.go
gammazero: I did not want to use a multiaddr, because I wanted it to be clear that this is a URL because the protocol is http/https.  I can change this, but it still needs to become a URL at some point.

engine/engine.go
gammazero: Since I already have the host, it seemed more direct to call `h.Addrs()` here.  In the other case of HTTP publisher, I do not want all the addresses, just the http address of the HTTP publisher.  For those reasons, I get the addresses part of the `AddrInfo` separately instead of getting the whole `AddrInfo`.

engine/engine.go
masih: That's fine. 

One could argue that if `ctx` is cancelled, then no announce should be called to begin with since there is no point calling an announce. 

Marking as non blocker.

masih (approved):

cmd/provider/daemon.go
```diff
@@ -125,6 +125,7 @@ func daemonCommand(cctx *cli.Context) error {
 	eng, err := engine.New(
 		engine.WithDatastore(ds),
 		engine.WithDataTransfer(dt),
+		engine.WithDirectAnnounce(cfg.DirectAnnounce.URLs...),
```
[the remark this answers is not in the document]
gammazero: > Two announcement protocols are supported, and used depending on the publisher kind

No, the announcement protocol used is not determined by the publisher kind. If there is an advertisement publisher, then _all available_ announce protocols are used. [here](https://github.com/filecoin-project/index-provider/pull/252/files#diff-2e6c4934f63ff9b712c2c346b33036af4724adf70b0801fff9b74f71b37fcd89L205)

Both an HTTP publisher or a dtsync publisher are announced using both gossipsub and http. 

I think this is better left as `WithDirectAnnounce` since we may also want to send direct announce messages over libp2p in the future.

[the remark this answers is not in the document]
gammazero: After changing the implementation to use multiaddrs, and since there is no need to support anything other than HTTP for sending direct announce messages,I have decided to go back to using URLs instead, for the following reasons:
- Communicates (or at least suggests) to the user that an HTTP endpoint is expected.
- Easier for users to to construct URL path if needed
- Does not require conversion from multihash to URL, and avoids any error cases associated with that.

PR #258, opened 2022-08-04
Support relative paths for car import

kaloyan-raev:
The CLI sends the car path to the daemon for import. The daemon may not resolve relative car paths if the CLI and daemon run in different working folders.

The proposed solution is the CLI to resolve the relative path to an absolute path before sending it to the daemon.


masih (approved):
Thank you 👍

PR #259, opened 2022-08-04
Calculate import cart key from the absolute car file path instead of raw path flag value

kaloyan-raev:
If the import car key is not specified by the user, it is currently calculated as a hash of the unclean car path.

So different paths to the same file would produce a different key and hence a different ad for the very same content.

For example, importing:

my.car
./my.car
../my.car
/abs/path/to/my.car

will result in a different ad although these paths point to the same file.

It would be better if the absolute car file path is used to determine the import car key.

### DIFF
--- a/cmd/provider/import.go
+++ b/cmd/provider/import.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"path"
 	"path/filepath"
 
 	"github.com/filecoin-project/index-provider/cardatatransfer"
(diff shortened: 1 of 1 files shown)


masih (commented):
Would it make sense to use the hash of absolute/clean car path instead? That should avoid conflict in a case where two car files have the same name but different content, located at different directories.

kaloyan-raev:
@masih Sure, makes sense. I made the change.

masih (approved):

PR #260, opened 2022-08-04
Add announce addresses and providerID to NotifyPut

gmelodie:
This makes `NotifyPut` more flexible, as it can optionally receive a `providerID string` and an `addresses []string` to announce on behalf of. This will be useful for the Estuary integration.


gmelodie:
Implemented by https://github.com/filecoin-project/index-provider/commit/ecf5e9c2ae971bd9035c1f21a91152e4a0bc6614

PR #261, opened 2022-08-05
Mirror an existing provider with optional entries remap and CDN

masih:
Implement the ability to periodically check the advertisement chain
published by an index provider an mirror it with optional remapping of
the original entries in HAMT or entry chunk.

Host the original ad chain in the mirror such that the mirror can also
act as a CDN for the original ad chain.

Implement a mirror testing environment that asserts various
configuration permutations function as expected.

Implement utilities to extract `MultihashIterator` from links stored by
an IPLD link system regardless of the underlying entries type.

### DIFF
--- a/mirror/mirror.go
+++ b/mirror/mirror.go
@@ -1,0 +1,396 @@
+package mirror
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"time"
+
+	dt "github.com/filecoin-project/go-data-transfer"
+	datatransfer "github.com/filecoin-project/go-data-transfer/impl"
+	dtnetwork "github.com/filecoin-project/go-data-transfer/network"
+	gstransport "github.com/filecoin-project/go-data-transfer/transport/graphsync"
+	"github.com/filecoin-project/go-legs"
+	"github.com/filecoin-project/go-legs/dtsync"
+	"github.com/filecoin-project/index-provider/engine/chunker"
+	"github.com/filecoin-project/storetheindex/api/v0/ingest/schema"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+	"github.com/ipfs/go-graphsync"
+	gsimpl "github.com/ipfs/go-graphsync/impl"
+	gsnet "github.com/ipfs/go-graphsync/network"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/linking"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+var log = logging.Logger("provider/mirror")
+
+// Mirror provides the ability to mirror the advertisement chain of an existing provider, with
+// options to restructure entries as EntryChunk chain or HAMT.
+//
+// Additionally, a mirror can also serve as a CDN for the original advertisement chain and its
+// entries. It exposes a GraphSync publisher endpoint from which ad chain can be synced.
+type Mirror struct {
+	*options
+	source  peer.AddrInfo
+	sub     *legs.Subscriber
+	pub     legs.Publisher
+	ls      ipld.LinkSystem
+	chunker *chunker.CachedEntriesChunker
+	cancel  context.CancelFunc
+}
+
+// New instantiates a new Mirror that mirrors ad chain from the given source provider.
+//
+// See: Mirror.Start, Mirror.Shutdown.
+func New(ctx context.Context, source peer.AddrInfo, o ...Option) (*Mirror, error) {
+	opts, err := newOptions(o...)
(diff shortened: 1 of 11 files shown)


mirror/mirror.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+var log = logging.Logger("provider/mirror")
+
+// Mirror provides the ability to mirror the advertisement chain of an existing provider, with
+// options to restructure entries as EntryChunk chain or HAMT.
+//
+// Additionally, a mirror can also serve as a CDN for the original advertisement chain and its
+// entries. It exposes a GraphSync
```
willscott, line 40:
```suggestion
// entries. It exposes a GraphSync publisher endpoint
```

mirror/mirror.go
```diff
@@ -0,0 +1,371 @@
...
+				log.Errorw("Failed to sync entries", "cid", entriesCid, "err", err)
+				return err
+			}
+			ad.Entries, err = m.remapEntries(ctx, ad.Entries)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	if err := ad.Sign(m.h.Peerstore().PrivKey(m.h.ID())); err != nil {
```
willscott, line 233: if we don't re-map, we could i guess retain the original signature?

willscott (commented):
Can you add a readme to the mirror directory explaining the semantics of how it can be applied:

it look like it acts like another provider with the same ads as an upstream - but it re-signs the advertisements as it's own identity, so there's an interaction between the provider of the content and the new publisher that we need to make clear in how this should be used.

mirror/mirror.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	return nil
+}
+
+func (m *Mirror) mirror(ctx context.Context, adCid cid.Cid) error {
+	ad, err := m.loadAd(ctx, adCid)
+	if err != nil {
+		return err
+	}
+	if err := ad.Validate(); err != nil {
```
gammazero, line 197: Is there any need to also check the signature on the ad? If a remote mirror is running, possibly mirroring specific providers, then that mirror may want to verify the ad signature.

mirror/mirror.go
```diff
@@ -0,0 +1,371 @@
...
+			if err != nil {
+				log.Errorw("Failed to sync entries", "cid", entriesCid, "err", err)
+				return err
+			}
+			ad.Entries, err = m.remapEntries(ctx, ad.Entries)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
gammazero, line 232: If `ad.IsRm` is true, then would it be a useful optimization to remember the ad's contextID and not mirror the entries for earlier ads with that contextID?

mirror/mirror_test.go
```diff
@@ -0,0 +1,208 @@
...
+				var ad *schema.Advertisement
+				ad, err = te.syncMirrorAd(ctx, gotMirroredHeadAdCid)
+				return err == nil && string(ad.ContextID) == "ad5"
+			}, testEventualTimeout, testCheckInterval, "err: %v", err)
+
+			// Read each individual mirrored ad and assert it is mirrored as expected.
+			te.requireAdChainMirroredRecursively(t, ctx, originalHeadAdCid, gotMirroredHeadAdCid)
+		})
+	}
+}
```
gammazero, line 208: Seems like mirroring a mirror is a possibility. Is that worth testing?

mirror/mirror.go
masih: Sure, this sounds like an optimisation and an option to enable disable it. Happy to take a look in a separate PR

mirror/mirror_test.go
masih: Sure. Is there a specific scenario you have in mind that could cause issues?

mirror/mirror.go
masih: Sure. I am curious what's the benefit in not resigning the ad? I can imagine a scenario where a mirror with persistent state is reconfigured such that the ad signature would differ. This would then mean some ads in the chain are signed by the original provider and some ads are signed by the mirror. 
Aside from that we need logic to determine when to resign and when not to resign. 

So I wonder is it not simpler to just resign ads by the mirror? and what's the benefit of not resigning considering a mirror always acts as a CDN for the original ad chain anyway?

mirror/mirror_test.go
gammazero: Cannot think of a specific situation. Maybe... check for correct behavior if 2nd mirror mirrors 1st mirror while 1st mirror is busy updating its content.  Make mirrors cannot mirror each other (causing a mirroring loop).

masih: > Make sure mirrors cannot mirror each other (causing a mirroring loop).

Not sure how we can guarantee this doesn't happen.

As for the former, I need to pick your brain to understand the scenario we are testing

mirror/mirror.go
masih: Changed the default behaviour to preserve the original signature whenever possible by default, with an option to always re-sign ads with the mirror's identity.

A mirror configured to not re-map the entries, would then mirror the original ad chain identically without any change. Tests are added to assert this. 

Please shout if there is anything else I might have missed.

mirror/mirror.go
masih: Left TODOs in code to add advertisement remapping as a feature. Thank you for pointing this out

willscott (approved):

PR #262, opened 2022-08-05
Added support fro multiple providers

ischasny:
This PR adds support for publishing on behalf of multiple providers. 
* `NotifyPut` and `NotifyRemove` APIs have been updated to take `peer.AddrInfo` and `peer.ID` respectively;
* If `nil` or an empty string have been passed in then the default configured provider will be assumed;
* Advertisements for all providers will appear on the same chain;
* There is no support for `RemoveAll` message yet that would clean up all ads for the given provider. Subject to the further PRs.


engine/engine.go
```diff
@@ -354,10 +355,10 @@ func (e *Engine) RegisterMultihashLister(mhl provider.MultihashLister) {
 // registered.
 //
 // See: Engine.RegisterMultihashLister, Engine.Publish.
-func (e *Engine) NotifyPut(ctx context.Context, contextID []byte, md metadata.Metadata) (cid.Cid, error) {
+func (e *Engine) NotifyPut(ctx context.Context, provider string, addrs []string, contextID []byte, md metadata.Metadata) (cid.Cid, error) {
```
willscott, line 358: consider using `peer.AddrInfo` as the way to talk about the provider+addrs, and `peer.ID` rather than string when just referring to the provider's identity.

ischasny: Makes sense, will change

masih, line 358: Take `peer.AddrInfo` instead of `string` as `provider` and `[]string` as `addrs`?

engine/engine.go
```diff
@@ -367,8 +368,8 @@ func (e *Engine) NotifyPut(ctx context.Context, contextID []byte, md metadata.Me
 // registered.
 //
 // See: Engine.RegisterMultihashLister, Engine.Publish.
-func (e *Engine) NotifyRemove(ctx context.Context, contextID []byte) (cid.Cid, error) {
-	return e.publishAdvForIndex(ctx, contextID, metadata.Metadata{}, true)
+func (e *Engine) NotifyRemove(ctx context.Context, provider string, contextID []byte) (cid.Cid, error) {
```
masih, line 371: Take `Peer.ID` instead of `string` as `provider`?

engine/engine.go
```diff
@@ -556,49 +562,121 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
 	return e.Publish(ctx, adv)
 }
 
-func (e *Engine) putKeyCidMap(ctx context.Context, contextID []byte, c cid.Cid) error {
+func (e *Engine) keyToCidKey(provider string, contextID []byte) datastore.Key {
```
masih, line 565: Would it make sense to take `peer.ID` wherever possible instead of `string` type?

engine/engine.go
```diff
@@ -556,49 +562,121 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+	return e.ds.Delete(ctx, e.keyToCidKey(provider, contextID))
+}
+
+func (e *Engine) deleteCidKeyMap(ctx context.Context, provider string, c cid.Cid) error {
+	return e.ds.Delete(ctx, e.cidToKeyKey(provider, c))
 }
 
-func (e *Engine) deleteCidKeyMap(ctx context.Context, c cid.Cid) error {
-	return e.ds.Delete(ctx, datastore.NewKey(cidToKeyMapPrefix+c.String()))
+type providerandcontext struct {
```
masih, line 631: `providerAndContext`

engine/engine.go
```diff
@@ -556,49 +562,121 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+	provider  string
+	contextID []byte
+}
+
+// returns all provider and context IDs associated with the given cid (a single cid can be advertised by different providers under same/different contexts)
```
masih, line 636:
```suggestion
// getAllCidKeyMap returns all provider and context IDs associated with the given cid.  A single cid can be advertised by different providers under the same/different contexts.
```

engine/engine_api_test.go
```diff
@@ -30,6 +31,21 @@ func (e *Engine) LinkSystem() *ipld.LinkSystem {
 	return &e.lsys
 }
 
+// Returns the engine's default provider ID, exposed for testing purposes only.
```
masih, line 34: Side note: godoc convention is to start the doc with the name of the function.

engine/engine_api_test.go
```diff
@@ -30,6 +31,21 @@ func (e *Engine) LinkSystem() *ipld.LinkSystem {
...
+
+// Returns the engine's default provider addresses, exposed for testing purposes only.
+func (e *Engine) ProviderAddrs() []string {
+	return e.retrievalAddrsAsString()
+}
+
+// Returns the engine's datastore, exposed for testing purposes only.
+func (e *Engine) Datastore() datastore.Datastore {
+	return e.ds
+}
```
masih, line 47: The tests are already in `engine` package which means they have access to all the engine fields including unexpored ones. 

I don't we need these.

engine/engine_test.go
```diff
@@ -347,6 +347,242 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+
+	wantContextID1 := []byte("fish")
+	wantContextID2 := []byte("bird")
+	subject.RegisterMultihashLister(func(ctx context.Context, p string, contextID []byte) (provider.MultihashIterator, error) {
+
+		if string(contextID) == string(wantContextID1) {
+			return provider.SliceMultihashIterator(mhs1), nil
+		} else if string(contextID) == string(wantContextID2) {
+			return provider.SliceMultihashIterator(mhs2), nil
+		}
```
masih, line 376: Would it make sense to make the conditions stronger by also asserting the passed provider ID  is the expected one for the given context ID?

engine/engine_test.go
```diff
@@ -347,6 +347,242 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+
+	gotLatestAdCid, ad, err = subject.GetLatestAdv(ctx)
+	require.NoError(t, err)
+	require.Equal(t, gotLatestAdCid, gotPutAdCid2)
+	require.Equal(t, ad.Provider, provider2id)
+	require.Equal(t, ad.Addresses, provider2Addrs)
+	require.Equal(t, ad.PreviousID.(cidlink.Link).Cid, gotPutAdCid1)
+}
+
+func TestEngine_WhenProviderIsEmpty_ShouldDefaultToTheProviderFromConfiguration(t *testing.T) {
```
masih, line 402: `TestEngine_UsesDefaultProviderWhenNoneGiven`

Totally cosmetic suggestion. See if you can find shorter names for the remaining tests.

engine/engine_test.go
```diff
@@ -347,6 +347,242 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
 	require.NotEqual(t, gotLatestAfterRmAdCid, gotLatestAdCid)
 }
 
+func TestEngine_WhenPublishForMultipleProviders_AdsShouldBeOnTheSameChain(t *testing.T) {
```
masih, line 350: `TestEngine_ProducesSingleChainForMultipleProviders`

engine/engine_test.go
```diff
@@ -347,6 +347,242 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+
+	gotPutAdCid1, err := subject.NotifyPut(ctx, provider1id, provider1Addrs, wantContextID, metadata.New(metadata.Bitswap{}))
+	require.NoError(t, err)
+	require.NotEqual(t, cid.Undef, gotPutAdCid1)
+	require.Equal(t, 1, subject.Chunker().Len())
+
+	gotPutAdCid2, err := subject.NotifyPut(ctx, provider2id, provider2Addrs, wantContextID, metadata.New(metadata.Bitswap{}))
+	require.NoError(t, err)
+	require.NotEqual(t, cid.Undef, gotPutAdCid2)
+	require.Equal(t, 1, subject.Chunker().Len())
```
masih, line 498: Would it make sense to also assert that ad entries link is the same across both ads?

engine/engine_test.go
```diff
@@ -347,6 +347,242 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	entriesCid := ad1.Entries.(cidlink.Link).Cid.String()
+	val, _ := ds.Get(ctx, datastore.NewKey("map/cidKey/"+entriesCid))
+	require.Equal(t, string(wantContextID), string(val))
+	val, _ = ds.Get(ctx, datastore.NewKey("map/keyCid/"+string(wantContextID)))
+	_, gotCid, _ := cid.CidFromBytes(val)
+	require.Equal(t, entriesCid, gotCid.String())
+	has, _ := ds.Has(ctx, datastore.NewKey("map/cidKey/"+subject.ProviderID()+"/"+entriesCid))
+	require.False(t, has)
+	has, _ = ds.Has(ctx, datastore.NewKey("map/keyCid/"+subject.ProviderID()+"/"+string(wantContextID)))
+	require.False(t, has)
```
masih, line 544: Would be good to generate a datastore from the `main` branch, and check it into code as `testdata`. We can then write a test that asserts all exported methods in engine work correctly given the initial datastore _and_ publishing new ads on top of the old one works gracefully.

Please remember to copy the `testdata` datastore into a temp directory and use the temp copy for testing since engine modifies it.

engine/example_test.go
```diff
@@ -64,7 +64,7 @@ func Example_advertiseHelloWorld() {
 
 	// Note that this example publishes an ad with bitswap metadata as an example.
 	// But it does not instantiate a bitswap server to serve retrievals.
-	adCid, err := engine.NotifyPut(context.Background(), []byte(sayHelloCtxID), md)
+	adCid, err := engine.NotifyPut(context.Background(), "", []string{}, []byte(sayHelloCtxID), md)
```
masih, line 67: `nil` suffices as empty slice. In go you can use `nil` for empty slice and calling `len` on it will return zero.

Ditto in other places.

engine/linksystem.go
```diff
@@ -83,30 +83,33 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
 			// If the link is not found, it means that the root link of the list has
 			// not been generated and we need to get the relationship between the cid
 			// received and the contextID so the lister knows how to
-			// regenerate the list of CIDs.
-			key, err := e.getCidKeyMap(ctx, c)
+			// regenerate the list of CIDs. There can be multiple provider + context IDs associated with a cid
+			// as index provider supports publishing on behalf of multiple providers.
+			keys, err := e.getAllCidKeyMap(ctx, c)
```
masih, line 88: `e.getAllCidKeyMap` would essentially iterate over the entire ad chain right? 
I am worried this would be expensive to do per read.

Would it make sense to:
- benchmark this, or
- think of maintaining some alternative reverse mapping that offers a cheaper/faster lookup?

engine/linksystem.go
```diff
@@ -83,30 +83,33 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
...
-			}
-
-			// Store the linked list entries in cache as we generate them.  We
-			// use the cache linksystem that stores entries in an in-memory
-			// datastore.
-			_, err = e.entriesChunker.Chunk(ctx, mhIter)
-			if err != nil {
-				log.Errorf("Error generating linked list from multihash lister: %s", err)
-				return nil, err
+			for _, key := range keys {
```
masih, line 94: If the CID for a set of entries is the the same, then by definition it means the multihashes they contained are also the same, since CID contains the digest of blocks it is associated to (with some caveats around hash function used) right?

If so then why entires are chunked for each of the `proivderandcontext` keys? 
Would doing this for any one of them once not be sufficient?

masih (changes requested):
One blocker on data typing: use `peer.ID` instead of string when referring to provider ID, and `peer.AddrInfo` instead of `string` and `[]string` when referring to provider ID and its addresses.

One question about the linksystem read opener lookup expensiveness, and regeneration of chunks for all keys.

Otherwise, nice work @ischasny 👍

engine/engine_api_test.go
ischasny: Well that's the thing - it wouldn't allow to just do engine.ds in the tests. I've tried before adding these methods.

engine/engine_api_test.go
ischasny: Fixed

engine/engine.go
ischasny: fixed

engine/engine.go
ischasny: fixed

engine/engine_test.go
ischasny: fixed

engine/engine_test.go
ischasny: good point, added to the condition

engine/engine_test.go
ischasny: fixed

engine/engine_test.go
ischasny: Good point, added

engine/linksystem.go
ischasny: Really great point, thanks! What that got me into thinking - is that we don't need to store cid -> context mappings for *all* providers. Storing it for *any* provider could be enough as chunking it for just one of them would be sufficient. Changed the code to maintain just one mapping and got rid of the expensive scan.

engine/engine.go
ischasny: Fixed

engine/engine.go
ischasny: Fixed

testutil/testutil.go
```diff
@@ -20,11 +24,19 @@ import (
...
+	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multihash"
 	"github.com/stretchr/testify/require"
 )
 
+func NewID(t testing.TB) peer.ID {
+	lp2p, _ := libp2p.New()
+	defer lp2p.Close()
```
masih, line 36: If all we want from a libp2p host is an ID, then we can generate random a id like [this](https://github.com/filecoin-project/storetheindex/blob/357c23af1b6929885a2ddf7329a1537e2449b6c5/config/init.go#L46). 

Cid V0 can also be a valid peer ID I think.

testutil/testutil.go
```diff
@@ -20,11 +24,19 @@ import (
...
 	"github.com/ipld/go-ipld-prime"
+	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multihash"
 	"github.com/stretchr/testify/require"
 )
 
+func NewID(t testing.TB) peer.ID {
+	lp2p, _ := libp2p.New()
```
masih, line 35: side-note: Never ignore `err` in test utilities specially when we have `t` to assert that no error has occurred.

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	require.NotEqual(t, cid.Undef, gotPutAdCid2)
+	require.Equal(t, 1, subject.Chunker().Len())
+	ad1, _ := subject.GetAdv(ctx, gotPutAdCid1)
+	ad2, _ := subject.GetAdv(ctx, gotPutAdCid2)
+
+	require.Equal(t, ad1.Entries, ad2.Entries)
+}
+
+func TestEngine_DatastoreBackwardsCompatibilityTest(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "DatastoreBackwardCompatibilityTest*")
```
masih, line 515: Use `t.TempDir()`. Testing would then clean it up for us.

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	if err != nil {
+		t.Fail()
+	}
+	defer os.RemoveAll(tempDir)
+	// copying testdata into the test dir
+	err = testutil.CopyDir(filepath.Join(testutil.ThisDir(t), "../testdata/datastore.ds"), tempDir)
```
masih, line 521: I am wondering if we can get away with having a single file that captures the ds. That would be nice, since we won't have to copy the whole directory.

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	if err != nil {
+		t.Fail()
```
masih, line 523: Use `require.NoError` consistently. Since these tests use testify library already, it makes sense to follow the existing pattern.

testutil/testutil.go
```diff
@@ -106,3 +118,63 @@ func SetupDataTransferOnHost(t *testing.T, h host.Host, store datastore.Batching
 	}
 	return dt
 }
+
+// CopyDir copies a whole directory recursively
+func CopyDir(src string, dst string) error {
```
masih, line 123: Take `testing.T` and assert no errors?

testutil/testutil.go
```diff
@@ -106,3 +118,63 @@ func SetupDataTransferOnHost(t *testing.T, h host.Host, store datastore.Batching
...
+
+	if dstfd, err = os.Create(dst); err != nil {
+		return err
+	}
+	defer dstfd.Close()
+
+	if _, err = io.Copy(dstfd, srcfd); err != nil {
+		return err
+	}
+	if srcinfo, err = os.Stat(src); err != nil {
```
masih, line 176: If we take `testing.T` we can also use `require.FileExists` and other utilities here.

masih (approved):
No blockers; it would be good to accept `nil` as `peer.AddrInfo` in `NotifyPut`. That should make the API a bit easier to use.

Left comments and suggestions. Please let me know if I can help expand on any.
It'd be awesome to clean up the git history before merge; rebase them to fewer (or maybe a single) commit with message explaining context, approach and consequences. 

Otherwise LGTM 👍

engine/engine.go
```diff
@@ -420,16 +423,16 @@ func (e *Engine) GetLatestAdv(ctx context.Context) (cid.Cid, *schema.Advertiseme
 	return latestAdCid, ad, nil
 }
 
-func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md metadata.Metadata, isRm bool) (cid.Cid, error) {
+func (e *Engine) publishAdvForIndex(ctx context.Context, p peer.ID, addrs []multiaddr.Multiaddr, contextID []byte, md metadata.Metadata, isRm bool) (cid.Cid, error) {
 	var err error
 	var cidsLnk cidlink.Link
 
-	log := log.With("contextID", base64.StdEncoding.EncodeToString(contextID))
+	log := log.With("providerID", p).With("contextID", base64.StdEncoding.EncodeToString(contextID))
```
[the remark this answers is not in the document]
ischasny: Good point - moved provider resolution up to the top

engine/engine.go
```diff
@@ -556,49 +569,116 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+func (e *Engine) keyToCidKey(provider peer.ID, contextID []byte) datastore.Key {
+	defaultProviderID := e.provider.ID
+	if provider == "" {
+		provider = defaultProviderID
+	}
+	if provider == defaultProviderID {
+		return datastore.NewKey(keyToCidMapPrefix + string(contextID))
+	} else {
+		return datastore.NewKey(keyToCidMapPrefix + string(contextID) + "/" + provider.String())
+	}
```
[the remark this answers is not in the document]
ischasny: applied

engine/engine.go
```diff
@@ -556,49 +569,116 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+func (e *Engine) keyToMetadataKey(provider peer.ID, contextID []byte) datastore.Key {
+	defaultProviderID := e.provider.ID
+	if provider == "" {
+		provider = defaultProviderID
+	}
+	if provider == defaultProviderID {
+		return datastore.NewKey(keyToMetadataMapPrefix + string(contextID))
+	} else {
+		return datastore.NewKey(keyToMetadataMapPrefix + string(contextID) + "/" + provider.String())
+	}
```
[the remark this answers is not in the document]
ischasny: applied

engine/engine.go
```diff
@@ -556,49 +569,116 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
 
-func (e *Engine) deleteKeyCidMap(ctx context.Context, contextID []byte) error {
-	return e.ds.Delete(ctx, datastore.NewKey(keyToCidMapPrefix+string(contextID)))
+func (e *Engine) deleteKeyCidMap(ctx context.Context, provider peer.ID, contextID []byte) error {
+	return e.ds.Delete(ctx, e.keyToCidKey(provider, contextID))
 }
 
-func (e *Engine) deleteCidKeyMap(ctx context.Context, c cid.Cid) error {
-	return e.ds.Delete(ctx, datastore.NewKey(cidToKeyMapPrefix+c.String()))
+func (e *Engine) deleteCidKeyMap(ctx context.Context, provider peer.ID, c cid.Cid) error {
```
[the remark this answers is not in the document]
ischasny: this is because neither cidToKey nor cidToProvider indexes have `providerId` in them. The former is a legacy index that didn't have a support for multiple providers. The latter stores mapping for *any* provider (whoever is the latest) as it's good enough to generate chunks when needed. I removed the extra parameter.

engine/engine.go
```diff
@@ -556,49 +569,116 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+	ContextID []byte
 }
 
-func (e *Engine) getCidKeyMap(ctx context.Context, c cid.Cid) ([]byte, error) {
-	return e.ds.Get(ctx, datastore.NewKey(cidToKeyMapPrefix+c.String()))
+func (e *Engine) getCidKeyMap(ctx context.Context, c cid.Cid) (*providerAndContext, error) {
+	// first see whether the mapping exists in the legacy index
+	val, err := e.ds.Get(ctx, e.cidToKeyKey(c))
+	if err == nil {
+		return &providerAndContext{Provider: peer.ID(""), ContextID: val}, nil
```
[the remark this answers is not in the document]
ischasny: applied

engine/engine_test.go
```diff
@@ -30,14 +34,17 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-multihash"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/time/rate"
 )
 
+var defaultProvider = peer.AddrInfo{ID: peer.ID(""), Addrs: nil}
```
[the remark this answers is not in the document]
ischasny: Fixed

mirror/mirror_env_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var defaultProvider = peer.AddrInfo{ID: peer.ID(""), Addrs: nil}
```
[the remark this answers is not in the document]
ischasny: Fixed

engine/linksystem_test.go
```diff
@@ -121,18 +124,18 @@ func Test_EvictedCachedEntriesChainIsRegeneratedGracefully(t *testing.T) {
 	ad2Mhs := testutil.RandomCids(t, rng, ad2MhCount)
 	require.NoError(t, err)
 
-	subject.RegisterMultihashLister(func(ctx context.Context, contextID []byte) (provider.MultihashIterator, error) {
+	subject.RegisterMultihashLister(func(ctx context.Context, p peer.ID, contextID []byte) (provider.MultihashIterator, error) {
 		strCtxID := string(contextID)
-		if strCtxID == string(ad1CtxID) {
+		if strCtxID == string(ad1CtxID) && p == peer.ID("") {
```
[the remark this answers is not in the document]
ischasny: Fixed

engine/engine.go
```diff
@@ -556,49 +569,116 @@ func (e *Engine) publishAdvForIndex(ctx context.Context, contextID []byte, md me
...
+		return nil, err
+	}
+	// trying to fetch this mapping from the new index
+	val, err = e.ds.Get(ctx, e.cidToProviderAndKeyKey(c))
+	if err != nil {
+		return nil, err
+	}
+
+	var pAndC providerAndContext
+	err = gob.NewDecoder(bytes.NewBuffer(val)).Decode(&pAndC)
```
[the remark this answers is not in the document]
ischasny: changed to json as discussed

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	_, err = subject.NotifyRemove(ctx, pID, []byte("star"))
+	require.NoError(t, err)
+
+	// try publishing for new provider
+	newPID := testutil.NewID(t)
+	_, err = subject.NotifyPut(ctx, peer.AddrInfo{ID: newPID}, []byte("has"), metadata.New(metadata.Bitswap{}))
+	require.NoError(t, err)
+}
+
+func verifyAd(t *testing.T, ctx context.Context, subject *engine.Engine, ad *schema.Advertisement, contextID []byte, provider string, addrs []string, entries string, isRm bool, prevId string) {
```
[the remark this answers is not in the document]
ischasny: makes sense - changed

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	require.Equal(t, ad.ContextID, contextID)
+	require.Equal(t, ad.Provider, provider)
+	require.Equal(t, ad.Addresses, addrs)
+	require.Equal(t, ad.Entries.(cidlink.Link).Cid.String(), entries)
+	require.Equal(t, ad.IsRm, isRm)
+	if prevId == "" {
```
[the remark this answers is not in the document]
ischasny: agreed, changed

engine/engine_test.go
```diff
@@ -347,6 +354,263 @@ func TestEngine_NotifyPutThenNotifyRemove(t *testing.T) {
...
+	} else {
+		require.Equal(t, ad.PreviousID.String(), prevId)
+	}
+
+	if !isRm {
+		chunk, err := subject.Chunker().GetRawCachedChunk(ctx, ad.Entries)
+		require.NoError(t, err)
+		require.NotNil(t, chunk)
+	} else {
+		println("")
```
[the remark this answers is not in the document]
ischasny: my bad, forgot to clean up, thanks!

engine/linksystem.go
```diff
@@ -95,7 +96,7 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
 			// deletes all indexes for the contextID in the removal
 			// advertisement.  Only if the removal had no contextID would the
 			// indexer ask for entry chunks to remove.
-			mhIter, err := e.mhLister(ctx, key)
+			mhIter, err := e.mhLister(ctx, key.Provider, key.ContextID)
```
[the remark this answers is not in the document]
ischasny: yes, good spot. I tweaked the behaviour a bit so that provider is resolved to default in the very beginning and hence no other functions should assume that the provider can be an empty string

engine/options.go
```diff
@@ -55,6 +55,8 @@ type (
 		// ID.
 		key crypto.PrivKey
 
+		// It's important to *not to change* this parameter when running against existing datastores. The reason for that is to maintain backward compatibility.
```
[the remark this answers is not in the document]
ischasny: Indeed, I think it's not supported - removed

engine/options.go
```diff
@@ -55,6 +55,8 @@ type (
...
+		// Older records from previous library versions aren't indexed by provider ID as it was assumed that there could be only one provider.
 		provider peer.AddrInfo
```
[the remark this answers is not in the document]
ischasny: added

interface.go
```diff
@@ -40,33 +41,35 @@ type Interface interface {
...
+	// is protocol dependant. The metadata must at least specify a protocol
 	// ID, but its data is optional.
 	//
-	// If both the contextID and metadata are the same as a previous call to
+	// If provider, contextID and metadata are the same as a previous call to
 	// NotifyPut, then ErrAlreadyAdvertised is returned.
 	//
 	// This function returns the ID of the advertisement published.
-	NotifyPut(ctx context.Context, contextID []byte, md metadata.Metadata) (cid.Cid, error)
+	NotifyPut(ctx context.Context, provider peer.AddrInfo, contextID []byte, md metadata.Metadata) (cid.Cid, error)
```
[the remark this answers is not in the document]
ischasny: Added

interface.go
```diff
@@ -40,33 +41,35 @@ type Interface interface {
...
+	// corresponded to the given provider and contextID are no longer available.  An advertisement
 	// is then generated, appended to the chain of advertisements and published
 	// onto the gossip pubsub channel.
-	// The given contextID must have previously been put via NotifyPut.
+	// The given provider and contextID tuple must have previously been put via NotifyPut.
 	// If not found ErrContextIDNotFound is returned.
 	//
 	// This function returns the ID of the advertisement published.
-	NotifyRemove(ctx context.Context, contextID []byte) (cid.Cid, error)
+	// TODO: add support for "delete all" for provider
```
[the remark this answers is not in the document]
ischasny: done

engine/engine.go
```diff
@@ -354,10 +357,10 @@ func (e *Engine) RegisterMultihashLister(mhl provider.MultihashLister) {
 // registered.
 //
 // See: Engine.RegisterMultihashLister, Engine.Publish.
-func (e *Engine) NotifyPut(ctx context.Context, contextID []byte, md metadata.Metadata) (cid.Cid, error) {
+func (e *Engine) NotifyPut(ctx context.Context, provider peer.AddrInfo, contextID []byte, md metadata.Metadata) (cid.Cid, error) {
```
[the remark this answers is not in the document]
ischasny: done

engine/linksystem_test.go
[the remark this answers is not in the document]
ischasny: fixed

testdata/datastore.ds/LOG
```diff
@@ -0,0 +1,8 @@
+=============== Aug 9, 2022 (BST) ===============
```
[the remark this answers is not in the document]
ischasny: indeed, removed both

interface.go
```diff
@@ -92,9 +95,10 @@ type MultihashIterator interface {
 	Next() (multihash.Multihash, error)
 }
 
-// MultihashLister lists the multihashes that correspond to a given contextID.
+// MultihashLister lists the multihashes that correspond to a given provider and contextIDs.
```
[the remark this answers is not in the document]
ischasny: fixed

testutil/testutil.go
ischasny: thanks it worked

testutil/testutil.go
ischasny: thanks, removed this block

engine/engine_test.go
ischasny: changed

engine/engine_test.go
ischasny: fixed

engine/engine_test.go
ischasny: two files are required unfortunately

testutil/testutil.go
ischasny: makes sense, fixed

testutil/testutil.go
ischasny: makes sense, changed

engine/engine.go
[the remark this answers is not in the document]
ischasny: applied

engine/engine.go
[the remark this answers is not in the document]
ischasny: applied

engine/engine.go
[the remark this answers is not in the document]
ischasny: Remove provider parameter. It's not used because it's not required for legacy index as there was only one provider neither it's required for the  new index as it stores *any* `cid -> (context, provider)` mapping as any is good enough to regenerate chunks

engine/engine.go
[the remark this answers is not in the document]
ischasny: applied and changed everywhere

PR #263, opened 2022-08-09
Implement the ability to run a mirror via `provider mirror` CLI

masih:
Add `provider mirror` command along with flags to set all possible
mirror options.

### DIFF
--- a/cmd/provider/mirror.go
+++ b/cmd/provider/mirror.go
@@ -1,0 +1,243 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"time"
+
+	"github.com/filecoin-project/index-provider/mirror"
+	leveldb "github.com/ipfs/go-ds-leveldb"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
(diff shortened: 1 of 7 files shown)


cmd/provider/mirror.go
```diff
@@ -0,0 +1,239 @@
...
+			Mirror.flags.entriesRecurLimit,
+			Mirror.flags.remapWithEntryChunkSize,
+			Mirror.flags.remapWithHamtHashFunc,
+			Mirror.flags.remapWithHamtBitWidth,
+			Mirror.flags.remapWithHamtBucketSize,
+			Mirror.flags.topic,
+			Mirror.flags.skipRemapOnEntriesTypeMatch,
+			Mirror.flags.alwaysReSignAds,
+		},
+		Before: func(cctx *cli.Context) error {
```
willscott, line 135: does this func need to be in `init`?

willscott (approved):

cmd/provider/mirror.go
masih: Not really. Happy to move it out if you think it makes the code harder to read.

willscott: I'd have these methods defined at the top level, rather than inline.

PR #273, opened 2022-09-24
add a catch-all unknown metadata codec

willscott:
for some cases where it's useful to have generic splitting/re-creation of provider metadata


metadata/unknown.go
```diff
@@ -0,0 +1,76 @@
...
+	// put the varint back in.
+	u.Payload = append(rl, buf...)
+
+	return int64(len(rl)) + bRead, nil
+}
+
+type rbr struct{ io.Reader }
+
+func (r rbr) ReadByte() (byte, error) {
+	var buf [1]byte
```
masih, line 67: Avoid unnecessary allocations by doing something like [this](https://github.com/ipld/go-car/blob/master/v2/internal/io/offset_read_seeker.go#L25).

metadata/unknown.go
```diff
@@ -0,0 +1,76 @@
...
+	bRead := int64(read)
+	if err != nil {
+		return int64(len(rl)) + bRead, err
+	}
+	if size != uint64(read) {
+		return int64(len(rl)) + bRead, fmt.Errorf("expected %d readable bytes but read %d", size, read)
+	}
+
+	// put the varint back in.
+	u.Payload = append(rl, buf...)
```
masih, line 59: Grow the buffer appropriately by sizing everything first to avoid allocs caused by slice expansion.

masih (commented):
Can I ask you to expand on why we need this? What specific usecase is this going to be applied to?

willscott:
Generic reframe / query code doesn't want to have to care about a specific set of known transports, e.g. if i get in arbitrary reframe protocols, i want to be able to pack them into the index-provider defined metadata format.
In the current metadata format there isn't a clean way to do that without doing something equivalent to this unknown transport object matching the interface in one's local code.

willscott:
Expanded this to refactor the interface a bit so that the additional codecs can actually be used.

Note that it's a breaking interface change now.

metadata/metadata.go
```diff
@@ -44,9 +45,37 @@ type (
...
+func init() {
+	Default = metadataContext{
+		protocols: make(map[multicodec.Code]func() Protocol),
+	}
+	Default.protocols[multicodec.TransportBitswap] = func() Protocol { return &Bitswap{} }
+	Default.protocols[multicodec.TransportGraphsyncFilecoinv1] = func() Protocol { return &GraphsyncFilecoinV1{} }
+}
+
+// WithProtocol dervies a new MetadataContext including the additional protocol mapping.
+func (mc *metadataContext) WithProtocol(id multicodec.Code, factory func() Protocol) metadataContext {
```
masih, line 64: * Unused?
* Exported function with unexported return type?

masih (approved):

metadata/metadata.go
willscott: the 'metadataContext' unexported  can still be used in the same way a `.Default` - it's hidden to prevent a client from making a malformed one directly

PR #275, opened 2022-09-28
Convert to metric unit before recording

masih:
Convert the mirror durations to millisecond to match the destination metric unit.


ischasny (approved):
LGTM

PR #278, opened 2022-10-08
Auto generate identity if one hasn't been provided in json config

ischasny:
The PR adds an ability to read a private key from a file if one hasn't been provided in the json config (similarly to `storetheindex`). If a key is neither in the config or a file - a new key is going to be generated and saved to the disk. 

Path to the file with private key is driven by `INDEXPROVIDER_PRIV_KEY_PATH ` environment variable. If neither config nor the var have been provided - the program will error out. 

This functionality has been requested by `bifrost-infra` in their efforts to automate `index-provider` deployments. CC @hsanjuan

### DIFF
--- a/cmd/provider/internal/config/identity_test.go
+++ b/cmd/provider/internal/config/identity_test.go
@@ -1,0 +1,54 @@
+package config
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	ic "github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDecodeOrCreate(t *testing.T) {
+	// first, creating an identity and saving the key to the file
+	identity := Identity{}
+	privKeyPath := filepath.Join(t.TempDir(), "privKey")
+	t.Setenv(PrivateKeyPathEnvVar, privKeyPath)
+
+	peerIDBefore, pkBefore, err := identity.DecodeOrCreate(io.Discard)
+	require.NoError(t, err)
+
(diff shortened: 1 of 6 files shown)


cmd/provider/internal/config/identity_test.go
```diff
@@ -0,0 +1,31 @@
...
+	// first, creating an identity and saving the key to the file
+	identity := Identity{}
+	privKeyPath := filepath.Join(t.TempDir(), "privKey")
+	t.Setenv(PrivateKeyPathEnvVar, privKeyPath)
+
+	peerIDBefore, pkBefore, err := identity.DecodeOrCreate(io.Discard)
+	require.NoError(t, err)
+
+	_, err = os.Stat(privKeyPath)
+	require.NoError(t, err)
```
masih, line 22: Assert that the identity is written out onto the file path and once decoded its ID matches the ID returned by `DecodeOrCreate`.

cmd/provider/internal/config/identity_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// verifying that once the file is there, the new identity is going to be read from it
+	identityFromPrivateKey := Identity{}
+	peerIDAfter, pkAfter, err := identityFromPrivateKey.DecodeOrCreate(io.Discard)
+	require.NoError(t, err)
+
+	require.Equal(t, pkBefore, pkAfter)
+	require.Equal(t, peerIDBefore, peerIDAfter)
```
masih, line 30: Assert that the file is not changed.

masih (approved):

cmd/provider/internal/config/identity_test.go
ischasny: added

PR #280, opened 2022-10-11
Add new branch to ECR publisher

ischasny:
Add ivan/add-ecr-public branch to ECR publisher config


masih (changes requested):
Please try adding this in `ivan/add-ecr-public` branch for testing purposes only and make sure to revert it after checking that the ECR build/push is successful.

PR #288, opened 2022-10-26
Stop removing individual expired reframe CIDs from datastore

ischasny:
Removing individual expired CIDs from datastore is too expensive. Switched to persisting the whole snapshot on clean ups instead, given that normally a large portion of CIDs expires in the same time.

### DIFF
--- a/reframe/listener.go
+++ b/reframe/listener.go
@@ -141,7 +141,7 @@ func New(ctx context.Context, engine provider.Interface,
 
 	// recording merged snapshot and cleaning up individual mappings from the datastore
 	if len(listener.cidQueue.listNodeByCid) > 0 {
-		listener.dsWrapper.recordTimestampsSnapshot(ctx, listener.cidQueue.getTimestampsSnapshot(), true)
+		listener.dsWrapper.recordTimestampsSnapshot(ctx, listener.cidQueue.getTimestampsSnapshot())
 	}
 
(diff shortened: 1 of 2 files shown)


masih:
I am curious; what is the runtime symptom of "removal bing too expensive"?

ischasny:
> I am curious; what is the runtime symptom of "removal bing too expensive"?

Web3 storage boxes get stuck for >1hr when a large chunk of Cids expires at the same time.

masih:
Sounds soar thank you

masih (approved):

reframe/listener.go
```diff
@@ -310,21 +310,13 @@ func (listener *ReframeListener) Provide(ctx context.Context, pr *client.Provide
...
-}
-
 // Revise logic here
-func (listener *ReframeListener) removeExpiredCids(ctx context.Context) error {
+func (listener *ReframeListener) removeExpiredCids(ctx context.Context) (bool, error) {
 	lastElem := listener.cidQueue.nodesLl.Back()
 	currentTime := time.Now()
 	chunksToRemove := make(map[string]*cidsChunk)
 	cidsToRemove := make(map[cid.Cid]struct{})
+	removedSomeCids := false
```
gammazero, line 319: nit: Since this is only used in the first `for` loop below, its declaration can be moved into the loop.  In go, if you declare a variable inside a loop, it will get initialized on each iteration.

gammazero (approved):
LGTM

PR #292, opened 2022-11-11
Add extended providers support

ischasny:
This PR adds support for [extended providers](https://github.com/filecoin-project/storetheindex/pull/804/files). It's a *programmatic* API only. CLI and configuration support will follow up if there going to be a need for those. 

In particular, the PR adds `xproviders.AdBuilder` class - a builder for constructing, signing and verifying an ad that can be then published using `Engine`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,33 @@ advertise content, see:
 
 * [`engine/example_test.go`](engine/example_test.go)
 
+#### Publishing ads with extended providers
+
+[Extended providers](https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md#extendedprovider) 
+field allows for specification of provider families, in cases where a provider operates multiple PeerIDs, perhaps 
+with different transport protocols between them, but over the same database of content. 
+
+Such ads can be composed manually or using a convenience builder `ExtendedProvidersAdBuilder`.
+```
+
+  adv, err := ep.NewExtendedProviderAdBuilder(providerID, priv, addrs). // the main ad's providerID, private key and addresses
+    WithContextID(contextID). // optional context id
+    WithMetadata(metadata). // optional metadata
+    WithOverride(override). // override flag, false by default
+    WithExtendedProviders(extendedProviders). // one or more extended providers to be included in the ad, represented by ExtendedProviderInfo struct
+    WithLastAdID(lastAdId). // cid of the last published ad, which is false by default
+    BuildAndSign()
+
+  if err != nil {
+    //...
+  }
+
+  engine.Publish(ctx, *adv)
+)
+```
+
+> Identity of the main provider will be added to the extended providers list automatically and should not be passed in explicitly
+
 ### `provider` CLI
 
 The `provider` CLI can be used to interact with a running daemon via the admin server to perform a
@@ -243,6 +270,7 @@ advertisement. The cache expansion is logged in `INFO` level at `provider/engine
 * [Indexer Node Design](https://www.notion.so/protocollabs/Indexer-Node-Design-4fb94471b6be4352b6849dc9b9527825)
 * [Providing data to a network indexer](https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md)
 * [`storetheindex`](https://github.com/filecoin-project/storetheindex): indexer node implementation
+* [`storetheindex` documentation](https://github.com/filecoin-project/storetheindex/blob/main/doc/)
 * [`go-indexer-core`](https://github.com/filecoin-project/go-indexer-core): Core index key-value
   store
 
--- a/engine/xproviders/xproviders.go
+++ b/engine/xproviders/xproviders.go
@@ -1,0 +1,164 @@
+package xproviders
+
+import (
+	"errors"
+
+	"github.com/filecoin-project/storetheindex/api/v0/ingest/schema"
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// AdBuilder contains fields required for building and signing of a new ad with Extended Providers
+type AdBuilder struct {
+	// providerID contains a peer ID of the main provider (the one from the body of the ad)
+	providerID string
+	// privKey contains a private key of the main provider
+	privKey crypto.PrivKey
+	// addrs contains addresses of the main provider
+	addrs []string
+	// providers contains providers' identities, keys, metadata and addresses that will be used as extended providers in the ad.
+	providers []Info
+	// contextID contains optional context id
+	contextID []byte
+	// contextID contains optional metadata
+	metadata []byte
+	// override contains override flag that is false by default
+	override bool
+	// lastAdID contains optional last ad cid which is cid.Undef by default
+	lastAdID cid.Cid
+}
+
+// Info contains information about extended provider.
+type Info struct {
+	// ID contains peer ID of the extended provider
+	ID string
+	// Metadata contains optional metadata of the extended provider
+	Metadata []byte
+	// Addrs contains a list of extended provider's addresses
+	Addrs []string
+	// Priv contains a provtae key of the extended provider
+	Priv crypto.PrivKey
+}
+
+// NewAdBuilder creates a new ExtendedProvidersAdBuilder
+func NewAdBuilder(providerID peer.ID, privKey crypto.PrivKey, addrs []ma.Multiaddr) *AdBuilder {
+	pub := &AdBuilder{
+		providerID: providerID.String(),
+		privKey:    privKey,
+		addrs:      multiaddrsToStrings(addrs),
+		providers:  []Info{},
+		lastAdID:   cid.Undef,
+	}
+	return pub
+}
+
+// WithContextID sets contextID
+func (pub *AdBuilder) WithContextID(contextID []byte) *AdBuilder {
+	pub.contextID = contextID
+	return pub
+}
+
+// WithMetadata sets metadata
+func (pub *AdBuilder) WithMetadata(md []byte) *AdBuilder {
+	pub.metadata = md
+	return pub
+}
+
+// WithOverride sets override
+func (pub *AdBuilder) WithOverride(override bool) *AdBuilder {
+	pub.override = override
+	return pub
+}
+
+// WithExtendedProviders sets extended providers
+func (pub *AdBuilder) WithExtendedProviders(eps ...Info) *AdBuilder {
+	pub.providers = append(pub.providers, eps...)
+	return pub
+}
+
+// WithLastAdID sets last ad cid
+func (pub *AdBuilder) WithLastAdID(lastAdID cid.Cid) *AdBuilder {
+	pub.lastAdID = lastAdID
+	return pub
+}
+
+// BuildAndSign verifies and  signs a new extended provider ad. After that it can be published using engine. Identity of the main provider will be appended to the
+// extended provider list automatically.
+func (pub *AdBuilder) BuildAndSign() (*schema.Advertisement, error) {
+	if len(pub.providers) == 0 {
+		return nil, errors.New("providers list is empty")
+	}
+
+	if len(pub.contextID) == 0 && pub.override {
+		return nil, errors.New("override is true for empty context")
+	}
+
+	adv := schema.Advertisement{
+		Provider:  pub.providerID,
+		Entries:   schema.NoEntries,
+		Addresses: pub.addrs,
+		ContextID: pub.contextID,
+		Metadata:  pub.metadata,
+		ExtendedProvider: &schema.ExtendedProvider{
+			Override: pub.override,
+		},
+	}
+
+	epMap := map[string]Info{}
+	for _, epInfo := range pub.providers {
+		adv.ExtendedProvider.Providers = append(adv.ExtendedProvider.Providers, schema.Provider{
+			ID:        epInfo.ID,
+			Addresses: epInfo.Addrs,
+			Metadata:  epInfo.Metadata,
+		})
+		epMap[epInfo.ID] = epInfo
+	}
+
+	// The main provider has to be on the extended list too
+	adv.ExtendedProvider.Providers = append(adv.ExtendedProvider.Providers, schema.Provider{
+		ID:        pub.providerID,
+		Addresses: pub.addrs,
+		Metadata:  pub.metadata,
+	})
+
+	if pub.lastAdID != cid.Undef {
+		prev := ipld.Link(cidlink.Link{Cid: pub.lastAdID})
+		adv.PreviousID = prev
+	}
+
+	err := adv.SignWithExtendedProviders(pub.privKey, func(provId string) (crypto.PrivKey, error) {
+		epInfo, ok := epMap[provId]
+		if !ok {
+			return nil, errors.New("unknown provider")
+		}
+		return epInfo.Priv, nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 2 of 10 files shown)


ischasny:
@gmelodie @hannahhoward this PR adds support for provider families. Please have a look whether it's sufficient for your usecase. Or would you need to have CLI support too?

go.mod
```diff
@@ -4,9 +4,9 @@ go 1.18
 
 require (
 	github.com/filecoin-project/go-data-transfer v1.15.2
-	github.com/filecoin-project/go-legs v0.4.16
+	github.com/filecoin-project/go-legs v0.4.17
 	github.com/filecoin-project/go-state-types v0.1.0
-	github.com/filecoin-project/storetheindex v0.4.28
+	github.com/filecoin-project/storetheindex v0.4.30-0.20221109112339-8dad1e315c7e
```
ischasny, line 9: I will update the reference before merging the PR.

engine/extended_provider/extended_providers_publisher.go
```diff
@@ -0,0 +1,155 @@
+package extended_provider
```
masih, line 1: What is the rationale for separating the package? Would be nice if we don't have to do this and the functionality simply extends the existing API?

engine/extended_provider/extended_providers_publisher.go
```diff
@@ -0,0 +1,155 @@
...
+		pub.providers[p.ID] = p
+	}
+}
+
+// Publish verifies, signs and publishes a new extended provider ad to the underlying engine. Identity of the main provider will be appended to the
+// extended provider list automatically and should not be passed in explicitly.
+//
+// - contextID and metadata fields are optional and should be used according to the specification
+// - providers list should contain only those providers whos infos have been registered in the publisher
+func (pub *ExtendedProvidersPublisher) Publish(ctx context.Context, contextID []byte, metadata []byte, override bool, providers ...peer.ID) (cid.Cid, error) {
```
masih, line 70: Wy not take Advertisement straight up? The number of arguments to this function seems to be a tad high.

If the idea is to assist with construction of `Advertisement` then maybe it would be more intuitive to abstract away `AdvertisementBuilder` instead of `Publish`.

engine/extended_provider/extended_providers_publisher.go
ischasny: My original idea was to add this API into the main interface, however if you think about it - there is nothing fundamentally new that has been added. I.e. the core engine functionality and the APIs that are exposed from it already are sufficient for the extended providers usecase too. 

Hence instead of modifying the existing APIs that would require at least 2 new parameters (override and ExtendedProviders) I decided to go with a layer on top of the existing engine. I also envision that the extended providers APIs are going to be used way less frequent comparing to the main `Publish` and `NotifyPut`. So modifying those to satisfy extended providers usecase was not an option either.

engine/extended_provider/extended_providers_publisher.go
ischasny: I thought about the builder. The issue is that assistance is required not just in the building, but in the signing too. Would your concern be alleviated if instead of passing in providers as a multiarg, I use array instead?

It's just +1 argument to what `NotifyPut` takes currently: 

```
NotifyPut(ctx context.Context, provider *peer.AddrInfo, contextID []byte, md metadata.Metadata) (cid.Cid, error)
```

README.md
```diff
@@ -153,6 +153,46 @@ advertise content, see:
...
+    []byte("some metadata"), // optional metadata that is included into the extended provider ad
+    "ABCSQP300ou/EsEnReN1BTvtjedhAjdAJGEydytYNwt6sAiM50lPfT64jA5P/RDf9o9IY0r+VtHdGgJYxyNAD0KJg=" // prviate key of the extended provider that is going to be used to sign its records in the ads
+  ), ....
+)
+```
+
+To publish an ad
+
+```
+publisher.Publish(
```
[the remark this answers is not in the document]
ischasny: ok, will modify it to be compilable

engine/extended_provider/extended_providers_publisher.go
ischasny: I thought about the builder while jogging and liked the idea :) Pushed the change in.

README.md
```diff
@@ -153,6 +153,33 @@ advertise content, see:
 
 * [`engine/example_test.go`](engine/example_test.go)
 
+#### Publishing ads with extended providers
+
+[Extended providers](https://github.com/filecoin-project/storetheindex/blob/main/doc/ingest.md#extendedprovider) 
+field allows for specification of provider families, in cases where a provider operates multiple PeerIDs, perhaps 
+with different transport protocols between them, but over the same database of content. 
+
+Such ads can be composed manually or using a conveniece builder `ExtendedProvidersAdBuilder`.
```
gammazero, line 162:
```suggestion
Such ads can be composed manually or using a convenience builder `ExtendedProvidersAdBuilder`.
```

README.md
```diff
@@ -153,6 +153,33 @@ advertise content, see:
...
+```
+
+  adv, err := ep.NewExtendedProviderAdBuilder(providerID, priv, addrs). // the main ad's providerID, private key and addresses
+    WithContextID(contextID). // optional context id
+    WithMetadata(metadata). // optional metadata
+    WithOverride(override). // override flag, false by default
+    WithExtendedProviders(extendedProviders). // one or more extended providers to be included into the ad, represnted by ExtendedProviderInfo struct
+    WithLastAdID(lastAdId). // cid of the last publisghed ad, which is false by default
```
gammazero, line 170:
```suggestion
    WithLastAdID(lastAdId). // cid of the last published ad, which is false by default
```

README.md
```diff
@@ -153,6 +153,33 @@ advertise content, see:
...
+with different transport protocols between them, but over the same database of content. 
+
+Such ads can be composed manually or using a conveniece builder `ExtendedProvidersAdBuilder`.
+```
+
+  adv, err := ep.NewExtendedProviderAdBuilder(providerID, priv, addrs). // the main ad's providerID, private key and addresses
+    WithContextID(contextID). // optional context id
+    WithMetadata(metadata). // optional metadata
+    WithOverride(override). // override flag, false by default
+    WithExtendedProviders(extendedProviders). // one or more extended providers to be included into the ad, represnted by ExtendedProviderInfo struct
```
gammazero, line 169:
```suggestion
    WithExtendedProviders(extendedProviders). // one or more extended providers to be included in the ad, represented by ExtendedProviderInfo struct
```

engine/extended_provider/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+	privKey crypto.PrivKey
+	// addrs contains addresses of the main provider
+	addrs []string
+	// providers contains providers' identities, keys, metadata and addresses that will be used as extended providers in the ad.
+	providers []ExtendedProviderInfo
+	// contextID contains optional context id
+	contextID []byte
+	// contextID contains optional metadata
+	metadata []byte
+	// override contains oiverride flag that is false by default
```
gammazero, line 29:
```suggestion
	// override contains override flag that is false by default
```

ischasny:
@masih ready for re-review

engine/extended_provider/extended_providers.go
```diff
@@ -0,0 +1,164 @@
+package extended_provider
```
gammazero, line 1: By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps. Please see:
- https://go.dev/blog/package-names
- https://go.dev/doc/effective_go#package-names

engine/extended_provider/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+	// contextID contains optional metadata
+	metadata []byte
+	// override contains oiverride flag that is false by default
+	override bool
+	// lastAdID contains optional last ad cid which is cid.Undef by default
+	lastAdID cid.Cid
+}
+
+// ExtendedProviderInfo contains information about extended provider.
+type ExtendedProviderInfo struct {
```
gammazero, line 36:
```suggestion
type Info struct {
```
https://github.com/golang/go/wiki/CodeReviewComments#package-names

gammazero (changes requested):
I must insist on package and type name changes. See comments.

engine/extended_provider/extended_providers_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func randomExtendedProvider(t *testing.T) (peer.ID, ep.ExtendedProviderInfo) {
+	rng := rand.New(rand.NewSource(time.Now().Unix()))
+	priv, _, providerID := testutil.GenerateKeysAndIdentity(t)
+	metadata := make([]byte, 0, 20)
+	_, err := rng.Read(metadata)
+	require.NoError(t, err)
+	addrs := make([]multiaddr.Multiaddr, 0, 2)
+	for i := 0; i < len(addrs); i++ {
+		s := fmt.Sprintf("/ip4/%d.%d.%d.%d/tcp/%d", rng.Int()%255, rng.Int()%255, rng.Int()%255, rng.Int()%255, rng.Int()%10000+1)
```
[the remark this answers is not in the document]
ischasny: Good spot

engine/extended_provider/extended_providers.go
ischasny: Changed

engine/extended_provider/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+	Priv crypto.PrivKey
+}
+
+// NewExtendedProviderAdBuilder creates a new ExtendedProvidersAdBuilder
+func NewExtendedProviderAdBuilder(providerID peer.ID, privKey crypto.PrivKey, addrs []ma.Multiaddr) *ExtendedProvidersAdBuilder {
+	pub := &ExtendedProvidersAdBuilder{
+		providerID: providerID.String(),
+		privKey:    privKey,
+		addrs:      multiaddrsToStrings(addrs),
+		providers:  []ExtendedProviderInfo{},
```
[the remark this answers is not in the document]
ischasny: Changed

engine/extended_provider/extended_providers.go
ischasny: Fixed

engine/extended_provider/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// ExtendedProvidersAdBuilder contains fields required for building and signing of a new ad with Extended Providers
+type ExtendedProvidersAdBuilder struct {
```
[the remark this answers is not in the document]
ischasny: Fixed and pushed, thanks for pointing out!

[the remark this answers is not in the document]
ischasny: Fixed and pushed, thanks for pointing out!

ischasny:
@gammazero fixes pushed please re-review

engine/extended_providers/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &adv, nil
+}
+
+// NewExtendedProviderInfo allows to create a new ExtendedProviderInfo providing type safety around its fields
+func NewExtendedProviderInfo(peerID peer.ID, priv crypto.PrivKey, metadata []byte, addrs []ma.Multiaddr) Info {
```
gammazero, line 149:
```suggestion
func NewInfo(peerID peer.ID, priv crypto.PrivKey, metadata []byte, addrs []ma.Multiaddr) Info {
```

engine/extended_providers/extended_providers.go
```diff
@@ -0,0 +1,164 @@
+package extendedproviders
```
gammazero, line 1: Better, but maybe `extproviders` or `xproviders` would be nice. Not a necessary change, but think about it.  If everyone if going to import it with an alias:
```go
import (
    xprov "github.com/filecoin-project/index-provider/engine/extended_provider"
)
```
Then maybe just change that package name to something shorter.


Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64.
-- Frrom: https://go.dev/doc/effective_go#package-names

gammazero (approved):
A couple more suggested changes, but not blocking. Do please change the directory name.

gmelodie:
Tagging @elijaharita

engine/extended_providers/extended_providers.go
ischasny: All great points, thanks! Changed and pushed.

engine/extended_providers/extended_providers.go
```diff
@@ -0,0 +1,164 @@
...
+// Info contains information about extended provider.
+type Info struct {
+	// ID contains peer ID of the extended provider
+	ID string
+	// Metadata contains optional metadata of the extended provider
+	Metadata []byte
+	// Addrs contains a list of extended provider's addresses
+	Addrs []string
+	// Priv contains a provtae key of the extended provider
+	Priv crypto.PrivKey
```
willscott, line 44:
```suggestion
	// Priv contains a private key of the extended provider
	Priv crypto.PrivKey
```

are there cases where the publishing process won't have the private keys of the whole family?

engine/xproviders/xproviders.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// AdBuilder contains fields required for building and signing of a new ad with Extended Providers
+type AdBuilder struct {
```
willscott, line 16: this looks very similar to the `extended_providers` package - do you mean to delete this one?

ischasny: Good spot - artifact after rebase onto the `main`. Cleaned up

engine/extended_providers/extended_providers.go
ischasny: I thought a lot about it. In our current usecase - allow a provider to scale up by adding new bitswap aliases on other libp2p hosts - the OG provider should have access too all XP keys. 

If going forwards we have such usecase when the main provider and it's XPs are in different security domains, then we can extend the protocol between two different `index-provider`s to be able to send partially signed ads to each other and assemble all required signatures that way.

willscott: I'd encourage checking with the providers to understand how this works with their current key management strategy - i don't think they have the keys being shared back currently, so if they're going to add a message protocol for comms to support that, we may need to jump to also supporting the partial signing

ischasny: Good point, noted.

hannahhoward (changes requested):
I may be missing it, but the missing feature for me is removing extended providers. While I suspect it will be a while before we do this on an cid by cid basis, from the beginning providers should be able to turn off bitswap after they enable it. If I'm missing something here, let me know.

ischasny:
> I may be missing it, but the missing feature for me is removing extended providers. While I suspect it will be a while before we do this on an cid by cid basis, from the beginning providers should be able to turn off bitswap after they enable it. If I'm missing something here, let me know.

Hey @hannahhoward , the idea is that you can publish an ad less the provider that needs to be removed. I.e. there is no a dedicated "remove" ad, just a new ad with an updated list. Would that work for you?

hannahhoward:
I think as long as you can do republish without the extended provider to remove, that works.

hannahhoward (approved):

PR #297, opened 2022-11-24
Update docs for engine and xproviders

ischasny:


masih (approved):
It would be great to also update godoc and add Go examples

PR #302, opened 2022-11-29
Configure address contained in HTTP announce messages

### LINKED ISSUE #299: Configured HTTP provider address not used
The configured HTTP provider address, `Ingest.HttpPublisher.ListenMultiaddr`, is not being used to config the provider engine. This means that only the default value,  `"/ip4/0.0.0.0/tcp/3104/http"`, is always used.

### LINKED ISSUE #300: Default publisher address causes bad address in HTTP announce message
When using the default config value, `"/ip4/0.0.0.0/tcp/3104/http"`, the address supplied in HTTP announce messages is always set to that value with the IP address of 0.0.0.0, instead of the address of the network interface the provider is bound to. This will generally work if the indexer and provider are on the same host, because 0.0.0.0 is replaced by the source address when connecting, but is incorrect.

### LINKED ISSUE #301: Need to configure HTTP announce address separate from publisher listen address
The address supplied in the HTTP announce message, optionally, needs to be configured separately from the address that the provider should listen on. This is necessary since a provider may have a different external address that indexers need to use to contact the provider.

If a separate HTTP announce message address is not configured, then use the address that the provider is listening on.

gammazero:
Configure the provider to listen on the configured `Ingest.HttpPublisher.ListenMultiaddr` address. Previously this was ignored.

The address configured in HTTP announce messages can be configured separately from the address that the provider should listen on. This is necessary since a provider may have a different external address that indexers need to use to contact the provider.

If the HTTP announce message address is not configured, then use the address of the interface that the provider listener is bound to. Previously, this was using "/ip4/0.0.0.0/tcp/3104/http" if the default config was used, instead of the interface address.

Fixes #299
Fixes #300
Fixes #301


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/filecoin-project/go-data-transfer v1.15.2
 	github.com/filecoin-project/go-state-types v0.1.0
-	github.com/filecoin-project/storetheindex v0.5.1
+	github.com/filecoin-project/storetheindex v0.5.2-0.20221129130444-841de0aa8b23
```
willscott, line 8: tag sti and merge against a released STI version, ideally

willscott (approved):

go.mod
gammazero: Will do that after approval of: https://github.com/filecoin-project/storetheindex/pull/1081

PR #306, opened 2023-01-05
Add page size configuration option

ischasny:
Propagate page size to reframe configuration so that it can be changed without redeploying the service


willscott (approved):

ischasny:
> will we need to convert the reframe provider to use the newer http delegated routing api soon?

Yes, true. Snapshot paging + http will be very beneficial. On my to-do list.

masih:
> Yes, true. Snapshot paging + http will be very beneficial. On my to-do list.

@ischasny  is there a github issue for this? thanks

ischasny:
Good point - [added](https://github.com/ipni/index-provider/issues/312)

PR #308, opened 2023-01-06
Improve reframe memory consumption

ischasny:
Fixed a bug when the same cids have been readvertised over and over again that has led to high disk and RAM usage. Removed some indexes that should improve RAM profile.

Fixes https://github.com/protocol/bifrost-infra/issues/2255

CC @gmasgras

### DIFF
--- a/reframe/chunker.go
+++ b/reframe/chunker.go
@@ -11,7 +11,6 @@ import (
 
 type chunker struct {
 	chunkByContextId map[string]*cidsChunk
-	chunkByCid       map[cid.Cid]*cidsChunk
 	currentChunk     *cidsChunk
 	chunkSizeFunc    func() int
 	nonceGen         func() []byte
@@ -35,7 +34,6 @@ func newChunker(chunkSizeFunc func() int, nonceGenFunc func() []byte) *chunker {
 	}
 	ch := &chunker{
 		chunkByContextId: make(map[string]*cidsChunk),
-		chunkByCid:       make(map[cid.Cid]*cidsChunk),
 		chunkSizeFunc:    chunkSizeFunc,
 		nonceGen:         nonceGenFunc,
 	}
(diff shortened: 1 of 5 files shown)


reframe/chunker.go
```diff
@@ -35,7 +34,6 @@ func newChunker(chunkSizeFunc func() int, nonceGenFunc func() []byte) *chunker {
 	}
 	ch := &chunker{
 		chunkByContextId: make(map[string]*cidsChunk),
-		chunkByCid:       make(map[cid.Cid]*cidsChunk),
```
ischasny, line 38: This index isn't required anymore as references to chunks get recorded in `cidNode`. Will allow to save some RAM on not needing to store the key twice.

reframe/listener.go
```diff
@@ -289,12 +291,14 @@ func (listener *ReframeListener) Provide(ctx context.Context, pr *client.Provide
 				//     * when currentChunk disappears between restarts as it doesn't get persisted until it's advertised
 				//     * when the same cid comes multiple times within the lifespan of the same chunk
 				//	   * after a error to generate a replacement chunk
-				err := listener.chunker.addCidToCurrentChunk(ctx, c, func(cc *cidsChunk) error {
-					return listener.notifyPutAndPersist(ctx, cc)
-				})
-				if err != nil {
-					log.Errorw("Error adding a cid to the current chunk. Continuing.", "cid", c, "err", err)
-					continue
+				if node.chunk == nil {
```
ischasny, line 294: That was the main bug that has led to the same cids been readvertised over and over again.

reframe/listener_test.go
```diff
@@ -663,7 +663,38 @@ func TestDoNotReadvertiseChunkIfAllCidsExpired(t *testing.T) {
 	require.Equal(t, []cid.Cid{testCid2}, reframelistener.GetExpiryQueue(ctx, listener))
 }
 
-func TestDoNoLoadRemovedChunksOnInitialisation(t *testing.T) {
+func TestDoNotReadvertiseTheSameCids(t *testing.T) {
```
ischasny, line 666: New test that verifies that the same CIDs will not be readvertised anymore.

willscott (approved):

PR #313, opened 2023-01-09
Fix reframe initialisation reporting

ischasny:
Chunks that are loaded during initialisation don't get stored in some in-memory indexes anymore because they have already ben indexed by the engine. Hence the index that have been used for reporting on initialisation would always report that zero chunks have been loaded.

### DIFF
--- a/reframe/listener.go
+++ b/reframe/listener.go
@@ -109,6 +109,9 @@ func New(ctx context.Context, engine provider.Interface,
 		func() int { return len(listener.chunker.currentChunk.Cids) },
 	)
 
+	// used for reporting only
+	uniqueChunks := make(map[string]struct{})
+
 	lister := &ReframeMultihashLister{
 		CidFetcher: func(contextID []byte) (map[cid.Cid]struct{}, error) {
 			ctxIdStr := contextIDToStr(contextID)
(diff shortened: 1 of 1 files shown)


reframe/listener.go
```diff
@@ -109,6 +109,9 @@ func New(ctx context.Context, engine provider.Interface,
 		func() int { return len(listener.chunker.currentChunk.Cids) },
 	)
 
+	// used for reporting only
+	uniqueChunks := make(map[string]struct{})
```
masih, line 113: This is going to store every unique context ID as [base64 encoded string](https://github.com/ipni/index-provider/blob/1d5d9beaea8b7c74dfc55ebf87b8f4a58733459a/reframe/listener.go#L511) just to report the length. Two question:

* Do we really need the count of unique chunks in the logs?
* If we really really do, would it make sense to hash the context id for a more predictable size growth of this map?

gammazero: Please just report how many chunks were loaded, by incrementing a counter, even if chunks are not unique.

ischasny: Agreed, I don't think this is really required anymore - closing it down.

Issue #314, opened 2023-01-16
extend `list ad` cli to print extended providers


willscott:
currently the extended providers fields are not printed on the CLI at all

PR #316, opened 2023-01-17
Update `provider ls` to print extended providers

masih:
Update the ad list command to print extended providers.

fixes https://github.com/ipni/index-provider/issues/314


willscott (approved):
fix #314

ischasny (approved):
Thanks for adding that

PR #320, opened 2023-01-18
Fix bug in xproviders builder

ischasny:
* Don't allow empty metadata or addresses in the extended provider body
* Don't add the main provider as extended if it's already on the list
* Don't add extended providers at all if non has been provided explicitly

Fixes https://github.com/ipni/index-provider/issues/318

CC @dirkmc

### DIFF
--- a/engine/xproviders/xproviders.go
+++ b/engine/xproviders/xproviders.go
@@ -105,21 +105,32 @@ func (pub *AdBuilder) BuildAndSign() (*schema.Advertisement, error) {
 	}
 
 	epMap := map[string]Info{}
-	for _, epInfo := range pub.providers {
-		adv.ExtendedProvider.Providers = append(adv.ExtendedProvider.Providers, schema.Provider{
-			ID:        epInfo.ID,
-			Addresses: epInfo.Addrs,
-			Metadata:  epInfo.Metadata,
-		})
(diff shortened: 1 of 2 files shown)


engine/xproviders/xproviders.go
```diff
@@ -105,21 +105,28 @@ func (pub *AdBuilder) BuildAndSign() (*schema.Advertisement, error) {
...
-		adv.ExtendedProvider.Providers = append(adv.ExtendedProvider.Providers, schema.Provider{
-			ID:        epInfo.ID,
-			Addresses: epInfo.Addrs,
-			Metadata:  epInfo.Metadata,
-		})
-		epMap[epInfo.ID] = epInfo
-	}
+	if len(pub.providers) != 0 {
+		for _, epInfo := range pub.providers {
+			if len(epInfo.Addrs) == 0 || len(epInfo.Metadata) == 0 {
```
masih, line 110: * `nil` Metadata in extended provider is allowed right? which should pick up the root metadata if specified.
* Check for ID not being empty, by calling `.Validate()` on it.

masih (changes requested):

engine/xproviders/xproviders.go
ischasny: Good catch - done

PR #321, opened 2023-01-18
Bump up version to v0.10.1

ischasny:


TorfinnOlsen:
Is the index provider update simply to accommodate the code changes we're testing for DHstore/reader privacy or was there something specifically changed on the index provider I should include in our updates? @ischasny

PR #332, opened 2023-02-10
Daemon needs to use RetrievalMultiaddrs from configuration

gammazero:
The provider daemon was not using the RetrievalMultiaddrs from the configuration file. This caused its advertisements to only contain the local addresses from the host, which would not be acceptable if the host is behind a gateway or NAT.

### DIFF
--- a/engine/options.go
+++ b/engine/options.go
@@ -315,9 +315,19 @@ func WithDatastore(ds datastore.Batching) Option {
 // indexing advertisement.
 // If unspecified, the libp2p host listen addresses are used.
 // See: WithHost.
-func WithRetrievalAddrs(addr ...multiaddr.Multiaddr) Option {
+func WithRetrievalAddrs(addrs ...string) Option {
 	return func(o *options) error {
-		o.provider.Addrs = addr
+		if len(addrs) != 0 {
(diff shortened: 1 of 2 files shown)


engine/options.go
```diff
@@ -315,9 +315,19 @@ func WithDatastore(ds datastore.Batching) Option {
...
+func WithRetrievalAddrs(addrs ...string) Option {
 	return func(o *options) error {
-		o.provider.Addrs = addr
+		if len(addrs) != 0 {
+			maddrs := make([]multiaddr.Multiaddr, len(addrs))
+			for i, a := range addrs {
+				var err error
+				maddrs[i], err = multiaddr.NewMultiaddr(a)
+				if err != nil {
+					return fmt.Errorf("Bad multiaddr %q: %w", a, err)
```
willscott: are we happy failing on any bad addr, or should we succeed as long as any is ok?

gammazero: Since this is a config file, configured by an admin, it seems that they would want this to fail on any bad address because they probably meant something else that could be important and would want to fix that mistake... so this forces them to do so.

Also, these are the addresses that go into advertisements, so if we ignore a bad one, then the correct version of it will be missing from all the advertisements until the admin notices it. ... and advertisements are generally permanent things.

PR #339, opened 2023-03-03
Updates go-data-transfer v2

hannahhoward:
# Goals

Update to new data transfer and associate libp2p deps

# Implementation

- had to also update storetheindex
- mostly updates cardatatransfer example repo. uses a new premade types repo that is propogating eventually to lassie and go-fil-markets to go-fil-markets to avoid copying over and over.
- I'm stuck on one failing test, and I think it has something to do with the "Topic" option being eliminated on the data transfer syncer in storetheindex. I could use some help with what to do here.


gammazero (approved):
Thank you for all this work.  I completed this PR by allowing adding back the option to specify an existing Topic, and setting up the senders that are used by the publishers to send announce messages. My change allows publishers (dtsync and http) to be able to send announce messages over both pubsub and HTTP.

gammazero:
Note: The 32-bit tests are failing, so these should be disabled.

rvagg:
overwriting go-test.yml is not ideal since it comes in from unified-ci, but if you pull in #331 to this I _think_ you sort out the issues that are block that, and you get to use the config file to disable 32-bit tests: https://github.com/protocol/.github#configuration

gammazero:
@rvagg Merged #331 into main (because go1.20 tests should fail until fixed), and rebased this PR on top of that.

gammazero:
@hannahhoward It looks like there may be a data race in go-statemachine: https://github.com/ipni/index-provider/actions/runs/4344356047/jobs/7589024318#step:28:1457

masih (approved):

PR #348, opened 2023-04-04
Upgrade to http delegated routing

ischasny:
Upgrade from reframe to http delegated rotuing as reframe isn't supported by kubo anymore. Datastore is backwards compatible.

Tested with Kubo v0.19.0.

Fixes https://github.com/ipni/index-provider/issues/312

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,16 +105,15 @@ provider import car -l http://localhost:3102 -i <path-to-car-file>
 
 Both CARv1 and CARv2 formats are supported. Index is regenerated on the fly if one is not present.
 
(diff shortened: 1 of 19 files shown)


willscott (approved):
thanks for bringing this up to date.

once we do this we'll want to update cluster nodes to use this newer integration, right?

ischasny:
> thanks for bringing this up to date.
> 
> once we do this we'll want to update cluster nodes to use this newer integration, right?

Yes, as long as they are on the latest version of Kubo. I'll sync up with Hector and George once this is merged. There are no functional changes here :)

PR #351, opened 2023-04-05
Add periodic ad flush

ischasny:
Currently the last chunk gets stuck in the delegated routing server until it's full. That might take a long time given that re-provide is relatively infrequent. This commit adds a worker that periodically checks the current chunk and publishes it if it's not empty and is old enough.

* Add a new configuration parameter `AdFlushFrequency` that defaults to 10 minutes
* Add a worker to delegated routing listener that periodically checks the current chunk

### DIFF
--- a/cmd/provider/internal/config/delegatedrouting.go
+++ b/cmd/provider/internal/config/delegatedrouting.go
@@ -11,6 +11,7 @@ const (
 	defaultDelegatedRoutingReadTimeout  = Duration(10 * time.Minute)
 	defaultDelegatedRoutingWriteTimeout = Duration(10 * time.Minute)
 	defaultDelegatedRoutingCidTtl       = Duration(24 * time.Hour)
+	defaultAdFlushFrequency             = Duration(10 * time.Minute)
 	defaultDelegatedRoutingChunkSize    = 1_000
 	defaultDelegatedRoutingSnapshotSize = 10_000
 	defaultPageSize                     = 5_000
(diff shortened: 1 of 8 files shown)


cmd/provider/internal/config/delegatedrouting.go
```diff
@@ -64,6 +69,9 @@ func (c *DelegatedRouting) PopulateDefaults() {
 	if c.CidTtl == 0 {
 		c.CidTtl = defaultDelegatedRoutingCidTtl
 	}
+	if c.AdFlushFrequency == 0 {
+		c.AdFlushFrequency = defaultAdFlushFrequency
```
willscott, line 73: should there be a way to not do the time-based flush?

delegatedrouting/listener.go
```diff
@@ -486,6 +496,42 @@ func contextIDToStr(contextID []byte) string {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			flushFunc()
+		}
+
+		t = time.NewTimer(listener.adFlushFrequency)
```
masih, line 526: Create `time.Ticker` once and reuse. This creates a ticker on every iteration.

masih (approved):

willscott (approved):
fine with this as-is, or you could use 0 to mean 'don't do this'

delegatedrouting/listener.go
ischasny: Good point - fixing

cmd/provider/internal/config/delegatedrouting.go
ischasny: Having zero to disable it sounds like a reasonable option. Let me add it.

PR #356, opened 2023-04-26
Some improvements to index provider delegated routing

ischasny:
* Chunks for removed advertisements aren't stored in the datastore anymore, that will allow to improve initialisation time significantly especially for long running nodes;
* Fixes a bug when chunks got paged incorrectly on initialisation. That resulted into some chunks not being loaded at all and some got duplicated;
* Fixes a bug when chunk hasn't been assigned to CIDs on initialisation which resulted into all CIDs being treated was "new".

### DIFF
--- a/delegatedrouting/chunker.go
+++ b/delegatedrouting/chunker.go
@@ -60,7 +60,7 @@ func (ch *chunker) removeChunk(chunk *cidsChunk) {
 }
 
 func (ch *chunker) setNewCurrentChunk() {
-	ch.currentChunk = &cidsChunk{Cids: make(map[cid.Cid]struct{}, ch.chunkSizeFunc())}
+	ch.currentChunk = &cidsChunk{Cids: make(map[cid.Cid]struct{}, ch.chunkSizeFunc()), Removed: false}
 	ch.currentChunkTime = time.Now()
 }
 
--- a/delegatedrouting/chunker.go
+++ b/delegatedrouting/chunker.go
@@ -21,6 +21,8 @@ type chunker struct {
 type cidsChunk struct {
 	ContextID []byte
 	Cids      map[cid.Cid]struct{}
+	// unused field left for backward compatibility purposes
+	Removed bool
 }
 
 func defaultNonceGen() []byte {
--- a/delegatedrouting/chunker.go
+++ b/delegatedrouting/chunker.go
@@ -21,7 +21,6 @@ type chunker struct {
 type cidsChunk struct {
 	ContextID []byte
 	Cids      map[cid.Cid]struct{}
-	Removed   bool
 }
 
 func defaultNonceGen() []byte {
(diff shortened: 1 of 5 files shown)


delegatedrouting/chunker.go
```diff
@@ -21,7 +21,6 @@ type chunker struct {
 type cidsChunk struct {
 	ContextID []byte
 	Cids      map[cid.Cid]struct{}
-	Removed   bool
```
ischasny, line 24: Remove flag is not required anymore as chunks for removed context ids simply get deleted from the datastore

delegatedrouting/ds_wrapper.go
```diff
@@ -44,18 +44,18 @@ func (dsw *dsWrapper) initialiseFromTheDatastore(ctx context.Context, cidImporte
...
 	for {
 		pageNum := offset / dsw.pageSize
-		log.Infof("Reading chunk page %d from the datastore. totalChunks=%d, removedChunks=%d, totalCids=%d", pageNum, pageNum*dsw.pageSize, removedChunks, totalCids)
+		log.Infof("Reading chunk page %d from the datastore. totalChunks=%d, totalCids=%d", pageNum, pageNum*dsw.pageSize, totalCids)
 
 		q := dsq.Query{
 			Prefix: chunkByContextIdIndexPrefix,
 			Offset: offset,
 			Limit:  dsw.pageSize,
+			Orders: []dsq.Order{dsq.OrderByKey{}},
```
ischasny, line 58: This fixes a bug with incorrect paging

delegatedrouting/listener.go
```diff
@@ -147,14 +147,21 @@ func New(ctx context.Context, engine provider.Interface,
...
 		now := time.Now()
-		// some timestamps might be missing in the case if the latest snapshot hasn't been persisted due to an error
-		// while some chunks containing those CIDs haven been persisted and sent out. In that case - backfilling the
-		// missing CIDs with the current timestamp. That is safe to do. Even if those CIDs have expired, they will still
-		// expire from the index-provider just at a later date.
 		for c := range chunk.Cids {
-			if listener.cidQueue.getNodeByCid(c) != nil {
+			// if the cid has already been registered - assign the chunk to it
+			if elem := listener.cidQueue.getNodeByCid(c); elem != nil {
+				node := elem.Value.(*cidNode)
```
ischasny, line 153: This fixes a bug when chunk shaven't been assigned to cids

delegatedrouting/chunker.go
willscott: is existing data potentially parsed into this struct?

willscott (approved):

delegatedrouting/chunker.go
ischasny: Existing datastores will have to be removed - they got too big (that's why I've been looking into this). As the vast majority of ads are removals, they eventually start constituting >90% of the datastore. The longer index-provider runs the bigger they get. However with the current improvement datastore will grow linearly with the number of CIDs. 

Removing datastore won't harm the index.

willscott: just making sure that we don't end up with de-serialization issues if an older data store is parsed by newer code

ischasny: Good point - will add the field back for backward compatibility

PR #358, opened 2023-05-12
feat: allow extended provider ad builder to set ad entries

alanshaw:
This PR adds a `WithEntries` method to the extended provider advertisement builder that allows it to set advertisement entries.

Fixes https://github.com/ipni/storetheindex/issues/1725


ischasny (approved):
Thank you for contributing

PR #362, opened 2023-05-17
Fix a bug when default provider identity hasn't been assumed in the l…

ischasny:
Cherry pick bug fix for boost.


masih (approved):
Thank you for getting this sorted so quickly @ischasny 🍺

PR #366, opened 2023-05-25
Remove commands that are now implemented in ipni-cli

gammazero:
Let's not maintain the same functionality in multiple places.


masih (approved):
🎉

PR #419, opened 2023-10-23
Test publish after remove for the context id

ischasny:
Verify that index-provider allows reusing the same context id. To do that one would need to call NotifyRemove followed by NotifyPut.


LexLuthr (approved):
This looks good. Now, we need to figure out a way to implement this in Boost for specific deals or maybe a manual command would make more sense.

ischasny:
> This looks good. Now, we need to figure out a way to implement this in Boost for specific deals or maybe a manual command would make more sense. 

Sounds good - let's discuss at the Colo.

PR #423, opened 2023-12-05
Add a hook to fix multihash mismatch errors

ischasny:
When a storage read error happens, the calling application can now invoke a custom piece of logic to fix that error in-flight.

### DIFF
--- a/engine/linksystem.go
+++ b/engine/linksystem.go
@@ -25,7 +25,9 @@ var (
 // Creates the main engine linksystem.
 func (e *Engine) mkLinkSystem() ipld.LinkSystem {
 	lsys := cidlink.DefaultLinkSystem()
-	lsys.StorageReadOpener = func(lctx ipld.LinkContext, lnk ipld.Link) (io.Reader, error) {
+
+	storageReadOpener := func(lctx ipld.LinkContext, lnk ipld.Link) (io.Reader, error) {
+
 		// If link corresponds to schema.NoEntries return error immediately.
 		if lnk == schema.NoEntries {
 			return nil, errNoEntries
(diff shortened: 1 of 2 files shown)


willscott (approved):

engine/linksystem.go
```diff
@@ -127,7 +127,11 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
 			}
 			if regeneratedLink == nil || !c.Equals(regeneratedLink.(cidlink.Link).Cid) {
 				log.Errorw("Regeneration of entries link from multihash iterator did not match the original link. Check that multihash iterator consistently returns the same entries for the same key.", "want", lnk, "got", regeneratedLink)
-				return nil, ErrEntriesLinkMismatch
+				if e.skipMultihashMismatchErros {
+					return nil, ipld.ErrNotExists{}
+				} else {
+					return nil, ErrEntriesLinkMismatch
```
gammazero, line 133: Is there any situation where `ErrEntriesLinkMismatch` should be returned?

At some point in the future, a callback can be called here that tries to fix the error. Maybe if that callback determines it can fix DB, but is not done yet, then `ErrEntriesLinkMismatch ` could be returned to have IPNI retry getting this ad at a later time when DB is fixed. Without such a callback, it seems like the only thing that can be done is to tell IPNI to skip this by returning `ipld.ErrNotExists{}`, and there should be no option.

ischasny: My idea was that what we are introducing with `ipld.ErrNotExists{}` is a hack to temporarily workaround the  issue that we are having. Once this is resolved, we should revert back to returning `ErrEntriesLinkMismatch`. WDYT?

gammazero: OK. That works as a way to work around after any alternatives have been exhausted.
👍

gammazero (approved):

PR #430, opened 2024-02-14
cmd: allow to set the retrieval peerID, not just the retrieval multiaddr

MichaelMure:
The config would allow us to specify a data retrieval multiaddr, but that is actually not enough as the default peerID (aka, the index-provider itself) is still used in the advertisements.

This PR add the necessary to also override the default peerID, which makes advertisements works as expected.

### DIFF
--- a/engine/options.go
+++ b/engine/options.go
@@ -372,6 +372,22 @@ func WithRetrievalAddrs(addrs ...string) Option {
 	}
 }
 
+// WithRetrievalPeerID sets the peerID of where to get the content corresponding to an indexing advertisement.
+// If unspecified, the libp2p host peer ID are used.
+// See: WithHost
+func WithRetrievalPeerID(peerID string) Option {
+	return func(o *options) error {
+		if len(peerID) != 0 {
+			pId, err := peer.Decode(peerID)
+			if err != nil {
+				return fmt.Errorf("bad peer ID %q: %w", peerID, err)
+			}
+			o.provider.ID = pId
+		}
+		return nil
+	}
+}
+
 func WithSyncPolicy(syncPolicy *policy.Policy) Option {
 	return func(o *options) error {
 		o.syncPolicy = syncPolicy
--- a/cmd/provider/daemon.go
+++ b/cmd/provider/daemon.go
@@ -160,6 +160,7 @@ func daemonCommand(cctx *cli.Context) error {
 		engine.WithHttpPublisherAnnounceAddr(cfg.Ingest.HttpPublisher.AnnounceMultiaddr),
 		engine.WithPubsubAnnounce(!cfg.DirectAnnounce.NoPubsubAnnounce),
 		engine.WithSyncPolicy(syncPolicy),
+		engine.WithRetrievalPeerID(cfg.ProviderServer.RetrievalPeerID),
 		engine.WithRetrievalAddrs(cfg.ProviderServer.RetrievalMultiaddrs...),
 	)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


engine/options.go
```diff
@@ -372,6 +372,22 @@ func WithRetrievalAddrs(addrs ...string) Option {
 	}
 }
 
+// WithRetrievalPeerID sets the peerID of where to get the content corresponding to an indexing advertisement.
+// If unspecified, the libp2p host peer ID are used.
+// See: WithHost
+func WithRetrievalPeerID(peerID string) Option {
```
masih, line 378: This is bound to provider libp2p host, since the identity is needed for signing of advertisements. 

Have you considered setting the engine host option instead?

MichaelMure: You mean just using `WithHost` and set the full retrieval host? Wouldn't that cause problems?

masih: It shouldn't. The identity of a provider is tied to what it uses[ to sign advertisements ](https://github.com/ipni/index-provider/blob/main/engine/options.go#L55-L59) and where chain is retrievable from. 
The reason for `RetrievalAddrs` option existing is that a libp2p node may have any public addr through DNS for example, that may not directly be known by it. This is similar to setting public hostname in a traditional HTTP server.

But the peer ID (or identity by extension) must be known to the host itself and must match the ID other peers expect to encounter when fetching ad chain from it.

EnchanterIO: For a bite more ctx, we have this PR deployed and it generates the following expected advertisements:

```
CID:          baguqeera7miiexnupzmlrusq2s5u2xd4ngqrsqormpjdtl6hkq73w2xqtkca
...
Signature: ✅ valid
Signed by: advertisement publisher
```

masih: @EnchanterIO What other options are set in the engine that is deployed?  I suspect the reason that works is that the peer ID set via option introduced by this PR matches the host identity?

EnchanterIO: It does not match the identity, contrary, the problem we try to address with this PR is that if `retrievalAddr` is not empty, the host id (identity of index provider) is inserted **into adv** - but we want our separate p2p-server bitswap provider peer id coming from this option to be inserted into adv instead.

masih: Got it thank you

In IPNI terminology they are referred to as "Provider" and "Publisher". Provider provides the actual downloadable content , and Publisher tells the network of what is available for download from where i.e. expose the advertisement chain.

IIUC what you'd like to be able to do is to use the engine as a publisher for providers that may reside elsewhere. Right? 

Related note: I recommend looking at Extended Providers in IPNI spec.

MichaelMure: > IIUC what you'd like to be able to do is to use the engine as a publisher for providers that may reside elsewhere. Right?
That is correct.

engine/options.go
```diff
@@ -372,6 +372,22 @@ func WithRetrievalAddrs(addrs ...string) Option {
...
+	return func(o *options) error {
+		if len(peerID) != 0 {
+			pId, err := peer.Decode(peerID)
+			if err != nil {
+				return fmt.Errorf("bad peer ID %q: %w", peerID, err)
+			}
+			o.provider.ID = pId
+		}
+		return nil
+	}
```
gammazero, line 388: This functionality is already available by using the `WithProvider` and `WithPrivateKey` options. The `WithProvider` option sets the peer and addresses for the provider to put in indexing advertisements. This is different than `WithHost` which sets the host that is publishing the advertisements (libp2p server).

Setting the provider ID should also be accompanied by a call to `WithPrivateKey`, to set the private key associated with the provider peer ID being set. This way the advertisements will be signed with the provider's key.

The only reason this PR currently works is because the indexer is currently configured to allow any ID to publish on behalf of a provider. This means the advertisement will be accepted if it has a valid signature from anyone. This configuration is long overdue for being locked down so that the indexer only accepts advertisements signed by the provider or from a specific list of delegated publishers.

gammazero (changes requested):

PR #441, opened 2024-05-08
feat: update boxo with new Provide interface

hacdias:
It is no longer possible to keep track of "snapshots" as the `Provide` function is called per announcement and it is not correlated to the HTTP request. I'm not sure how to proceed here.


lidel:
cc https://github.com/ipni/index-provider/issues/403

PR #453, opened 2024-09-17
Read schema type hint from sync request

gammazero:
Read the schema type hint from the IPNI sync request header, and if provided, compare it against the type of data retrieved.


engine/linksystem.go
```diff
@@ -52,14 +61,24 @@ func (e *Engine) mkLinkSystem() ipld.LinkSystem {
...
 				return bytes.NewBuffer(val), nil
 			}
 			log.Debugw("Retrieved non-advertisement object from datastore", "cid", c, "size", len(val))
 		}
 
+		if cidSchemaType != "" && cidSchemaType != ipnisync.CidSchemaEntryChunk {
+			log.Errorf("Asked for %s when expecting %s", cidSchemaType, ipnisync.CidSchemaEntryChunk)
+			return nil, errors.New("requested unexpected data type")
+		}
+
```
willscott, line 81: lets log these but not return an error?

if we got the right data that matches the cid, we should continue, not error

